PR #2, opened 2019-01-30
JRE-1166 [jdk11] Handle the case when fonts are installed into use…

mfilippov:
…r registry key. This is the default behaviour since Windows 10 1809.


forantar:
Thank you, looks fine to me (however, assigned to Dmitry who has expertise in fonts).

PR #3, opened 2019-03-01
Adopt fix to match patch submitted into Oracle

mfilippov:


mfilippov:
@JB-Dmitry ping

JB-Dmitry:
I'll merge it when it will be accepted in OpenJDK. Who knows, maybe some more changes will be required.

PR #7, opened 2019-06-20
README: Bintray links now point to the files tab

devnoname120:


avu:
Thanks for the update but we finally decided to switch to openjdk11. We'll update README with appropriate links.

PR #8, opened 2019-07-24
Update 11.0.4

bell-sw:
Here we come with a jdk-11.0.4-ga update from OpenJDK [all changes, including security fixes].
There was a couple of merge conflict resolved, see them in a separate commits in the beginning of the commit list.
The update was tested on MacOS, Windows, Linux - all OK.
Sincerely yours, BellSoft.


avu:
Thanks for the update!

PR #10, opened 2019-08-06
Lazy font family population

denofevil:
Avoid calling expensive native getWidth on font when creating font family, load styles only when they would be used.

### DIFF
--- a/src/java.desktop/share/classes/sun/font/FontFamily.java
+++ b/src/java.desktop/share/classes/sun/font/FontFamily.java
@@ -28,7 +28,9 @@ package sun.font;
 import java.io.File;
 import java.awt.Font;
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Collection;
+import java.util.List;
 import java.util.HashMap;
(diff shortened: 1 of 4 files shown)


ignatov:
Don't forget to add 
```
package sun.font;

public interface FontWithDerivedItalic {
    Font2D createItalic();
}
```

src/java.desktop/share/classes/sun/font/FontFamily.java
```diff
@@ -257,7 +262,36 @@ public void setFont(Font2D font, int style) {
...
+            doSetFont(font, style);
+            return;
+        }
+        if (fontSequence == null) {
+            fontSequence = new ArrayList<FontAndStyle>();
+        }
+        fontSequence.add(new FontAndStyle(font, style));
+    }
+
+    private void ensureFontsLoaded() {
```
JB-Dmitry: This method and other relevant places should be synchronized, unless you've checked that such a synchronization is guaranteed by the calling methods.

PR #18, opened 2020-05-17
Border style changes

tatakir:
Implemented a fully-native border rendering algorithm of top border

### DIFF
--- a/src/java.desktop/windows/native/libawt/windows/awt_Frame.cpp
+++ b/src/java.desktop/windows/native/libawt/windows/awt_Frame.cpp
@@ -1839,17 +1839,11 @@ MsgRouting AwtFrame::WmNcCalcSize(BOOL wParam, LPNCCALCSIZE_PARAMS lpncsp, LRESU
     if (!wParam || !HasCustomDecoration()) {
         return AwtWindow::WmNcCalcSize(wParam, lpncsp, retVal);
     }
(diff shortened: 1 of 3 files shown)


ForNeVeR:
This is part of implementation of [IDEA-241054](https://youtrack.jetbrains.com/issue/IDEA-241054), I'll manage it from now.

ForNeVeR:
One issue we've found is that the changes are always enabled. But they should only be enabled when custom window frame is enabled.

src/java.desktop/windows/native/libawt/windows/awt_Frame.cpp
```diff
@@ -40,6 +40,15 @@
 #include <sun_awt_windows_WEmbeddedFrame.h>
 #include <sun_awt_windows_WEmbeddedFramePeer.h>
 
+#pragma comment (lib,"Gdiplus.lib")
+#pragma comment(lib, "UXTheme")
+#pragma comment (lib, "Dwmapi.lib")
+#pragma comment (lib, "Shcore.lib")
+
+#define DWMWA_USE_IMMERSIVE_DARK_MODE_BEFORE_20H1 19
+#ifndef DWMWA_USE_IMMERSIVE_DARK_MODE
```
ForNeVeR, line 49: This seems to be finally published officially in https://learn.microsoft.com/en-us/windows/win32/api/dwmapi/ne-dwmapi-dwmwindowattribute

PR #20, opened 2020-05-26
Dcevm

skybber:
This pull request contains:
- formal code modifications such as comments improvements, std Oracle documentation formatting etc.
- opens for `com.sun.beans.util` package, that contains reflection cache. It is necessary to flush it after redefinition if reflection is used.
- improved separation of standard code from DCEVM, frequently used by checking flag `AllowEnhancedClassRedefinition`. Many asserts from standard JVM were comment out in the past, all of them are back now for `AllowEnhancedClassRedefinition=false` 
- dcevm has own adjusting method `adjust_method_entries_dcevm` in `ResolvedMethodTable`
- improved support for lambdas, it fixes https://github.com/TravaOpenJDK/trava-jdk-11-dcevm/issues/21
- GC run after redefinition is skipped now if instance is not changed (level of  method or class modification). GC run is skipped in this way in dcevm7 and dcevm8. I've introduced full GC run after redefinition in dcevm11 since dcevm was more stable with it. I'm returning back to original code now since our long-term test does not crash now. Another reason for it is fact, that without full GC the redefinition is faster.
- Fix "no original bytecode found" - it fixes crashes if IDE deploys erroneous (partial) code
- Replace deleted method with Universe::throw_no_such_method_error - it is inspired by standard JDK code, that introduced it recently.

### DIFF
--- a/src/hotspot/share/classfile/classFileParser.cpp
+++ b/src/hotspot/share/classfile/classFileParser.cpp
@@ -954,6 +954,7 @@ void ClassFileParser::parse_interfaces(const ClassFileStream* const stream,
                                                   CHECK);
       }
 
+      // (DCEVM) pick newest
       interf = (Klass *) maybe_newest(interf);
 
       if (!interf->is_interface()) {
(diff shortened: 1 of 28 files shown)


artspb:
Thanks! I've started a build and will switch to this version as soon as it's ready.

PR #26, opened 2020-07-07
JBR-2479: API for getting JIT compiler state

Alex2772:
Implemented API for acquiring JIT compiler state from Java code


bulenkov:
Please move all public APIs to com.jetbrains.*

PR #27, opened 2020-07-09
JBR-2479: API for getting JIT compiler state (moved to com.jetbrains)

Alex2772:
Implemented API for acquiring JIT compiler state from Java code. Moved implementation to com.jetbrains.management due to license restrictions

### DIFF
--- a/src/java.management/share/native/libmanagement/JitStateImpl.c
+++ b/src/java.management/share/native/libmanagement/JitStateImpl.c
@@ -1,0 +1,32 @@
+/*
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+#include "management.h"
+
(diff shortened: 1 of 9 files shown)


src/java.management/share/native/libmanagement/JitStateImpl.c
```diff
@@ -0,0 +1,32 @@
...
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+#include "management.h"
+
+JNIEXPORT jboolean JNICALL
+Java_com_jetbrains_management_JitState_isCompilationEnabled
+        (JNIEnv *env, jobject dummy)
+{
+    return jmm_interface->GetBoolAttribute(env, JMM_COMPILATION_ENABLED);
```
donnerpeter, line 25: Do you need jmm_interface here? Can you call CompileBroker directly?

Alex2772: CompileBroker is a C++ interface, while the code above is written in C. I can export `CompileBroker::get_compilation_activity_mode` using new "C"-externed function but why should am I create a new function if Java provides a more elegant and standardized way to bind C++ interface to C code.

In addition to, this module is C only, I cannot create C++ source file there. I cannot include compileBroker.hpp, because libmanagement module does not have that header in scope.

donnerpeter: Can the native method implementations be moved into some module where interfacing with C++ is easier? Involving JMM flags feels superfluous.

Alex2772: Creating own modules in JVM is hard and looks superfluous too.

donnerpeter: Maybe reuse some existing one?

Alex2772: I'm not arguing. Which module should I use?

donnerpeter: If only I knew. Are there modules with C++ that see CompileBroker and can contain JNI bindings?

Alex2772: java.base, java.desktop, jdk.pack and hotspot contains C++ code. Maybe I should have put JNI function into the hotspot since it can see CompileBroker but AFAIK it does not contain Java code at all.

donnerpeter:
"hotspot" looks closest, but adding Java classes there probably isn't nice
indeed. OK, let's leave it as is for now.

PR #55, opened 2021-02-10
JBR-3068: pass default button localization if not overridden

ForNeVeR:


forantar (approved):

forantar:
Thank you, pushed it: c790bf3

PR #58, opened 2021-03-22
RIDER-57780: reload state of native file dialogs each time

ForNeVeR:
To work around [RIDER-57780](https://youtrack.jetbrains.com/issue/RIDER-57780), we need to switch common file dialogs on an off at runtime sometimes. This PR allows us to do it (JBR will reload the state of the `useCommonItemDialog` property every time a new dialog is instantiated).

This shouldn't affect any other products than Rider, since nobody else uses the native file dialogs of any kind.


forantar:
Thank you, pushed.

PR #78, opened 2021-08-29
Fix IDEA-196116: Don't touch the active keyboard layout on input method activation / deactivation

ForNeVeR:
This is a fix for [IDEA-196116](https://youtrack.jetbrains.com/issue/IDEA-196116): essentially, I propose to stop touching the active layout on frame / input method activation/deactivation, and let the OS to do all the work.

### DIFF
--- a/src/java.desktop/windows/classes/sun/awt/windows/WInputMethod.java
+++ b/src/java.desktop/windows/classes/sun/awt/windows/WInputMethod.java
@@ -168,10 +168,6 @@ final class WInputMethod extends InputMethodAdapter
 
     @Override
     public boolean setLocale(Locale lang) {
-        return setLocale(lang, false);
-    }
-
-    private boolean setLocale(Locale lang, boolean onActivate) {
         Locale[] available = WInputMethodDescriptor.getAvailableLocalesInternal();
         for (int i = 0; i < available.length; i++) {
             Locale locale = available[i];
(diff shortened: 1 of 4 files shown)


NikitkoCent:
Looks like the patch breaks `Let me use a different input method for each app window` functionality: when you create multiple windows (reproduced with `JFrame`s) and this option is enabled, switching keyboard layout in one window also does this with other window(s).

Repro:
```java
import javax.swing.*;
import java.awt.*;

class Reproducer {
    public static void main(String[] args) {
        SwingUtilities.invokeLater(() -> {
            final JFrame jFrame = new JFrame("Reproducer Window 1");

            jFrame.setDefaultCloseOperation(WindowConstants.DISPOSE_ON_CLOSE);
            jFrame.setSize(new Dimension(500,500));

            final JButton jButton = new JButton("Show modal dialog");
            jButton.addActionListener(ignored -> {
                JOptionPane.showMessageDialog(jFrame, "I am a modal dialog");
            });
            jFrame.add(jButton);

            jFrame.setVisible(true);


            final JFrame jFrame2 = new JFrame("Reproducer Window 2");
            jFrame2.setDefaultCloseOperation(WindowConstants.DISPOSE_ON_CLOSE);
            jFrame2.setSize(400, 400);
            jFrame2.setVisible(true);
        });
    }
}
```

ForNeVeR:
@NikitkoCent, surprisingly (for me at least), it is how it's intended to work.

Checked on other multi-window applications (Chrome, VSCode, Visual Studio): they all have the same behavior as you described.

"Let me use a different input method for each app window" should have emphasis on "app": they mean different input method _per app instance_ (usually, per process), and not _per window_.

NikitkoCent:
PR has been merged into master (see 2f772fd1a2fad666b732b1078aff8380b128e0c6 commit) with the following changes:
* Commit message of the fix is modified (because issue's id has been changed);
* `NULL` at `awt_InputMethod.cpp:350` is replaced by `0`.

PR #81, opened 2021-09-12
Store unrecognized VM options into the java.vm.unrecognized.option

dsamersoff:
This patch

1) Change default behavior to ignore unrecognized options
(+IgnoreUnrecognizedVMOptions)

2) Store all ignored values to:

"java.vm.unrecognized.option"

as

":+BogusOption01:+BogusOption02"

3) Explicitly set -IgnoreUnrecognizedVMOptions for couple of tests, that rely on default behaviour


dsamersoff:
Something is wrong with github, it keeps adding commits to this PR. Closing

PR #83, opened 2021-09-16
Jbr dev vmops

dsamersoff:
This patch:

1.  Change default behavior to ignore unrecognized options (+IgnoreUnrecognizedVMOptions)

2. Store all ignored values to: "java.vm.unrecognized.options" as "+BogusOption01 +BogusOption02" (space separated), see testcase

3. Explicitly set -IgnoreUnrecognizedVMOptions for couple of tests, that rely on default behavior

PS: Default separator for multiple values in property provided by os::path_separator() function, this behavior changed by second commit. The separator hard coded to " " (space)

### DIFF
--- a/src/hotspot/share/runtime/arguments.cpp
+++ b/src/hotspot/share/runtime/arguments.cpp
@@ -118,6 +118,8 @@ bool Arguments::_has_jimage = false;
 
 char* Arguments::_ext_dirs = NULL;
 
+GrowableArray<const char *> *Arguments::_unrecognized_vm_options = NULL;
+
 bool PathString::set_value(const char *value) {
   if (_value != NULL) {
     FreeHeap(_value);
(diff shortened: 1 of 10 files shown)


mkartashev:
Space as a separator is no good either:
```
$ java -XX:+IgnoreUnrecognizedVMOptions -XX:+Use -XX:+Use\ G1\ GC Unrecognized                                                                                                                           
java.vm.unrecognized.options=+Use +Use G1 GC
```
I think the safest separator characters would be '\000', it's the only one that can't possibly be inside a program argument. And we don't need `java.vm.unrecognized.options` to be human-readable as a whole anyway, it's always going to be split before use.
Even the line break ('\n') can appear in the java argument (although you have to try even harder than with space to achieve that).

If '\000' won't work, I guess we'll need some other way to pass the list of options from the VM into Java.

avu:
Actually, we still can use space as a delimiter but we need to escape the spaces inside the options first.

mkartashev:
> Actually, we still can use space as a delimiter but we need to escape the spaces inside the options first.

That is a possibility. But we'll then have to actually parse the value of `java.vm.unrecognized.options` instead of simply doing `String.split()`.

mkartashev (commented):
Looks good to me.

PR #103, opened 2021-12-10
JBR-4107: A11y: macOS - wrong frame position if window is not on primary screen

VeselovAlex:
When application window is moved to another screen with different vertical resolution, a11y frames are placed into wrong position on screen. It looks like we are trying to transform coordinates using window main screen as an origin, but it should be a primary screen.


forantar (approved):

forantar:
Thanks for the fix!

Issue #84, opened 2021-09-16
JBR very slow compared to OpenJDK


lamka02sk:
Hi, I was reporting this issue already in YouTrack, and mentioning this multiple times, but nobody cares :(
The JBR on Linux is extremely slow compared to OpenJDK. About twice as slow. The problem is, OpenJDK doesn't really work well with 2021.2 IDEs, so I had to downgrade and I will be probably stuck on versions 2021.1 forever since this issue with JBR has not been adressed for over a year now.
I will gladly sent you any information or do any testing to help you finally fix this issue.
For benchmark, I locked my CPU to static frequency (1.4GHz on all cores) and did this small benchmark:
OpenJDK 11 - PHPStorm startup with medium project: 23 seconds (stopped when code editor and file tree loaded)
JBR bundled with 2021.1 - PHPStorm startup of the same project: 45 seconds

All of my devices have this problem and all are running Linux (Kubuntu 21.04, Ubuntu 20.04 LTS KDE Neon and Ubuntu 18.04 LTS KDE Neon)

Thank you for your help

System info:
```
Operating System: Kubuntu 21.04
KDE Plasma Version: 5.22.4
KDE Frameworks Version: 5.85.0
Qt Version: 5.15.2
Kernel Version: 5.14.2 (64-bit)
Graphics Platform: X11
Processors: 8 × AMD Ryzen 7 4700U with Radeon Graphics
Memory: 14.5 GiB of RAM
Graphics Processor: AMD RENOIR
```

bulenkov:
@lamka02sk Thank you for the report. We will start an investigation asap and will let you know about the progress. Could you please go to PhpStorm main menu Help -> About and click on Copy button, and then share the result?

lamka02sk:
Thank you for your response. Here is the output you requested with removed sensitive info. When I am on my main PC later today, I will post also config from PHPStorm 2021.2 here.

```
PhpStorm 2021.1.4
Build #PS-211.7628.25, built on June 30, 2021
Runtime version: 11.0.12+7-post-Debian-2deb10u1 amd64
VM: OpenJDK 64-Bit Server VM by Debian
Linux 5.14.2
GC: G1 Young Generation, G1 Old Generation
Memory: 1600M
Cores: 8
[...]

PR #121, opened 2022-03-22
Cache results of font-config pattern search for reuse

osialr:
This is a potential fix for #84.  libfontmanager invokes the fontconfig matcher very often, especially on editor font-size. The implementation caches the results of the calls.

On my system, each call to match takes between 10-20 ms, with a few >100ms.  Kubuntu + libreoffice exacerbates the problem, because it installs a lot of fonts and rules.  But even on a minimal install of ubuntu, this should help because each call takes 1-10ms.

See [IDEA-287246](https://youtrack.jetbrains.com/issue/IDEA-287246) for more debug diagnostics.


osialr:
@avu I have signed the CLA.

You can change NUM_CACHED_VALUES to 0 to disable the caching mechanism to compare the performance.  I choose the value of 8 empirically based on what smoothed out the UI sufficiently without frequent eviction.

I didn't try tuning the Pattern to FcFontMatch.  Maybe there is an improved pattern that quickly retrieves the required values.

avu (approved):

avu:
@osialr the changes look good, I've run commit testing and pushed them into JBR11 master

osialr:
Great.  I'm glad to help.

avu:
@osialr thank you very much for your help!

PR #129, opened 2022-04-13
Mkartash/wayland dev

mkartashev:


src/java.desktop/unix/classes/sun/awt/wl/WLToolkit.java
```diff
@@ -119,29 +125,28 @@ public FramePeer createFrame(Frame target) {
...
+                /* Events have been successfully read and handled */
                 SunToolkit.postEvent(AppContext.getAppContext(),
                         new PeerEvent(this, () -> {
-                            dispatchEvents();
+                            dispatchEventsOnEDT();
                             synchronized (WLToolkit.this) {
                                 WLToolkit.this.notifyAll();
                             }
                         }, PeerEvent.ULTIMATE_PRIORITY_EVENT));
-            } else if (result == READ_RESULT_NOT_STARTED) {
```
JB-Dmitry, line 130: This branch was needed to avoid spamming EDT with unnecessary PeerEvent-s. AFAIR, without it, CPU usage for AWT process went to 100% immediately. This happened when some Wayland events arrived to default queue, but haven't been yet dispatched by EDT. Looks like now this problem should be back.

src/java.desktop/unix/classes/sun/awt/wl/WLToolkit.java
```diff
@@ -119,29 +125,28 @@ public FramePeer createFrame(Frame target) {
...
+            if (result == READ_RESULT_ERROR) {
+                break;
+            } else if (result == READ_RESULT_FINISHED_WITH_EVENTS) {
+                /* Events have been successfully read and handled */
                 SunToolkit.postEvent(AppContext.getAppContext(),
                         new PeerEvent(this, () -> {
-                            dispatchEvents();
+                            dispatchEventsOnEDT();
                             synchronized (WLToolkit.this) {
                                 WLToolkit.this.notifyAll();
```
JB-Dmitry, line 137: With your version of the code, this `notifyAll` doesn't do anything. But see the comment below.

JB-Dmitry (changes requested):

src/java.desktop/unix/classes/sun/awt/wl/WLToolkit.java
mkartashev: What I don't understand here is why `poll` with a 50ms timeout was not enough to prevent 100% CPU usage. I'll take a closer look at what happens now wrt CPU usage.

mkartashev: Just checked: a JFrame app CPU usage is currently around 1-1.3% over the course of several minutes.

JB-Dmitry: Suppose a poll resulted in events being queued, we post a `PeerEvent` to process them. Suppose that event isn't processed at once, e.g. due to EDT being busy at the moment. So we execute `readEvents` again, and it immediately exits after `wl_display_prepare_read` check, as the queue is still non-empty. Now we post a `PeerEvent` again, and a busy loop continues like that until the events from Wayland queue will be drained, without even getting to the new poll.

mkartashev: Finally understood, thanks for the detailed explanation. Will revert this part of the commit.

src/java.desktop/unix/classes/sun/awt/wl/WLToolkit.java
```diff
@@ -73,12 +73,43 @@
 import sun.awt.datatransfer.DataTransferer;
 import sun.util.logging.PlatformLogger;
 
+/**
+ * Events handling: the WLToolkit class creates a thread named "AWT-Wayland" where all the Wayland
+ * event handlers are invoked. For "system" events that are not meant to trigger any user code,
+ * this is the only thread where they can be handled. For the rest, such as mouse click events,
+ * the Wayland handlers are supposed to "transfer" themselves to AWT-EventThread by means of
+ * SunToolkit.postEvent(). See the implementation of run() method for more comments.
+ */
```
JB-Dmitry, line 82: This comment doesn't describe the current situation correctly - not all event handlers are invoked on toolkit thread. Actually, only Robot event handlers are. Other ones are invoked on EDT, and `postEvent` only delays their execution, not changing the thread.

PR #141, opened 2022-05-07
Tag AWT windows as sRGB surfaces

rsauciuc:
OSX assumes windows are in the display's color space by default.
Java2D assumes its output in sRGB, which means the sRGB colors
are interpreted by the OS as being in the display's color space.
This means that unless the display is set to use the sRGB color
space, the final result on screen will look incorrect.

Issue: http://issuetracker.google.com/208508346
Change-Id: Ie4c8dd2e41ba4f92db0c59e94825512ea93c6553

### DIFF
--- a/src/java.desktop/macosx/native/libawt_lwawt/awt/AWTWindow.m
+++ b/src/java.desktop/macosx/native/libawt_lwawt/awt/AWTWindow.m
@@ -463,6 +463,9 @@ AWT_ASSERT_APPKIT_THREAD;
     if (self.nsWindow == nil) return nil; // no hope either
     [self.nsWindow release]; // the property retains the object already
 
+    // Tell the system we have an sRGB backing store
+    [self.nsWindow setColorSpace: [NSColorSpace sRGBColorSpace]];
+
     self.isEnabled = YES;
     self.isMinimizing = NO;
     self.javaPlatformWindow = platformWindow;


avu:
@rsauciuc did you suggest this change to OpenJDK ? I wonder why such a simple fix has not been integrated in the main line.

romainguy:
The "problem" with this change in OpenJDK is that it could potentially impact existing apps. macOS windows are by default in the display color space and apps must manually tag them to be sRGB or properly convert colors. While this fix is safe in Android Studio/IntelliJ (since those apps assume sRGB colors), it might not be for *all* JVM apps (even though it would be for *most* if not almost all…). Ideally OpenJDK would have an AWT API to set the color space on the window itself :/

rock3r:
Just FYI, I am seeing similar colour space problems under Windows (tested with AS Electric Eel canary 2 and IJ 222 EAP).

Also, this PR should fix issues such as https://youtrack.jetbrains.com/issue/IDEA-264808/Colours-Scheme-colours-are-displayed-wrong-different

avu:
> The "problem" with this change in OpenJDK is that it could potentially impact existing apps. macOS windows are by default in the display color space and apps must manually tag them to be sRGB or properly convert colors. While this fix is safe in Android Studio/IntelliJ (since those apps assume sRGB colors), it might not be for _all_ JVM apps (even though it would be for _most_ if not almost all…). Ideally OpenJDK would have an AWT API to set the color space on the window itself :/

I see. Recently we've introduced so-called [JBR api](https://github.com/JetBrains/JetBrainsRuntime/tree/main/src/jetbrains.api) to provide our IDEs with such additional features and in the same time keep our fork compliant with JCK. I think it would be better to provide this functionality via JBR api.

PR #143, opened 2022-05-09
Fix view hierarchy restoration for full-screen mode on macOS

manu-unter:
In the previous version of the code, we didn't correctly restore the translation of resize masks into constraints for the `titlebar` view, which mostly worked but had one edge case where it broke:
If the window already takes up the full width of the screen even before entering full-screen mode, then the constraints aren't updated correctly and there will be no window controls in the hidden title bar that appears when moving the mouse to the top of the screen.

See more details in this ticket: https://youtrack.jetbrains.com/issue/FL-11420/Exit-full-screen-buttons-disappears-in-full-screen-mode

I also merged the handlers for the transparent title bar into the existing full screen handlers - I must have missed those before.


avu (approved):
Looks good

PR #146, opened 2022-05-11
Master17

alexander-lobas:
For frame with several tabs when show only frame (without any dialogs and etc) do ignore `orderFront()' when frame activated. If user do click in inactive tab we have several `orderAboveSiblings()` and order front do cancel user choose.


avu (approved):
Looks good

PR #149, opened 2022-05-17
Make bash scripts safer

VladRassokhin:
That would help detect build problems earlier than having empty artifacts like in https://youtrack.jetbrains.com/issue/JBR-4451

### DIFF
--- a/jb/project/docker/mkdocker_aarch64.sh
+++ b/jb/project/docker/mkdocker_aarch64.sh
@@ -1,4 +1,7 @@
-#!/bin/bash -x
+#!/bin/bash
+
+set -euo pipefail
+set -x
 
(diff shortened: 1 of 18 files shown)


vprovodin:
at the exit we should do some actions (see do_exit in common.sh). Will it be done in case of set -euo pipefail ?

VladRassokhin:
Hopefully it would work. Would be better to replace such function with with `trap`, but not now.

vprovodin:
integrated, squashed into one commit

NikitkoCent:
@vprovodin Should we also forward-port this to `main` / `jbr-dev` branch?

vprovodin:
yes we should. It is still in progress.

vprovodin:
ported into `main`

PR #153, opened 2022-06-07
FL-11598 Fix accidental reset of window configuration on macOS

manu-unter:
This also aligns the handling of the style mask with how the other settings are applied: It uses a client property now.

The property is now also applied directly on peer initialization so that there is no need to re-apply, and the window directly appears in the correct configuration.

The issue that previously occurred was that enabling custom window decoration through the JBR API wasn't enough to keep it reliably there. It would reset everytime we applied style bits again because the internal stored style bits didn't contain the correct configuration regarding `WINDOW_FULL_CONTENT`, `WINDOW_TITLE_VISIBLE`, and `WINDOW_TRANSPARENT_TITLE_BAR`. This lead to the behavior as seen in the recording.

With these changes, the custom window decoration is now correctly applied through a custom property and thus also stored in the internal style bits.

https://user-images.githubusercontent.com/4516798/172323054-d701d6a1-cc52-46ea-8d26-aea402a2bcef.mov


avu (approved):
Looks good

PR #166, opened 2022-08-24
Global hotkeys on Windows

tsarn:
**NOTE: this is a solution to a take home assignment for SWE position at Jetbrains, treat it accordingly :)**

This PR adds a `GlobalHotkeyService` that allows registering callbacks to hotkeys that work without having the application's window in focus. It's implemented only for Windows, but can probably be extended to other platforms as well. It uses ServiceProvider to load platform-specific implementations.

Hotkeys are implemented by spawning a separate thread that specifically handles WM_HOTKEY events. This provides independency from a specific UI framework and event loop. It's probably not the cleanest possible solution, but it works quite well. You can see my [screenshot tool](https://github.com/tsarn/screenshot-tool) that demonstrates the use of these global hotkeys.

Another way would be to register hotkeys using AWT's window handles, process their events in AWT's event loop and treat hotkey events similar to all the other UI events.


tsarn:
@mkartashev @NikitkoCent I've just noticed that a recent rebase on the main branch broke this PR a bit, hope that's okay with you :)

NikitkoCent:
Great! Wait a letter 👍

PR #167, opened 2022-09-05
Jbr17

alexander-lobas:
[JBR-4787](https://youtrack.jetbrains.com/issue/JBR-4787) Rounded corners for native Window on Windows 11


src/java.desktop/share/classes/java/awt/Window.java
```diff
@@ -4080,6 +4080,32 @@ private interface MacOS {
         }
     }
 
+    private static class RoundedCornersManager {
```
YaaZ, line 4083: I will update JBR API to accept list of target mappings instead of single one, that will get us rid of all this boilerplate with internal services. And of course this will allow us to remove `isAvailable` method, as it will be built into JBR API

alexander-lobas: Okey

src/java.desktop/share/classes/java/awt/Window.java
```diff
@@ -4080,6 +4080,32 @@ private interface MacOS {
...
+        boolean isAvailable() {
+            return MacOS.INSTANCE != null || Win.INSTANCE != null;
+        }
+
+        void setRoundedCorners(Window window, Object params) {
```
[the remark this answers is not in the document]
alexander-lobas: While there is no implementation for Linux, I do not fully know this signature.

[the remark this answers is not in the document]
YaaZ: As soon as we push this into the API, we'll not be able to change the signature.
Maybe we can accept float here? Then we can use it as is on macOS and add few enum-like constants to use on Windows. And then depending on how it works on Linux, we can use it as a continuous or enum-like value too.

test/jdk/jb/java/awt/Window/TestSetRoundedCorners.java
```diff
@@ -0,0 +1,179 @@
...
+        validateColor(51, 51, Color.GREEN, "0");
+        validateColor(51, 99, Color.GREEN, "0");
+        validateColor(99, 51, Color.GREEN, "0");
+        validateColor(99, 99, Color.GREEN, "0");
+
+        runSwing(() -> testFrame.setVisible(false));
+        robot.delay(DELAY);
+
+        runSwing(() -> {
+            JBR.getRoundedCornersManager().setRoundedCorners(testFrame, roundParams);
```
YaaZ, line 93: These tests are run under JTreg, so JBR API will not be available here, and that would fail to compile in Teamcity. I think it needs to be moved to `test/jdk/jb/java/api/frontend`. See `FileDialogTest` and `JBRApiTest`, they use a script to bootstrap real JBR API for tests.

src/java.desktop/share/classes/java/awt/Window.java
YaaZ: https://jetbrains.team/p/jbre/reviews/35/timeline

test/jdk/jb/java/awt/Window/TestSetRoundedCorners.java
alexander-lobas: Moved

test/jdk/jb/java/api/frontend/TestSetRoundedCorners.java
```diff
@@ -0,0 +1,177 @@
...
+import java.awt.*;
+import java.awt.image.BufferedImage;
+import java.lang.reflect.InvocationTargetException;
+
+/**
+ * @test
+ * @summary JBR-4563 Rounded corners for native Window on Mac OS
+ * @summary JBR-4787 Rounded corners for native Window on Windows 11
+ * @author Alexander Lobas
+ * @run main TestSetRoundedCorners
```
YaaZ, line 36: It needs `@run shell build.sh` before the main test run to build API classes

alexander-lobas: done

PR #184, opened 2022-11-07
8284675: "jpackage.exe" creates application launcher without Windows …

manu-unter:
…Application Manfiest

Reviewed-by: erikj

This is required to fix https://youtrack.jetbrains.com/issue/FL-14792/while-using-multiple-screen-pixels-seems-distorted


NikitkoCent:
As far as I see, it's already in jbr17 (see the commit 716c07ef007dee83b106d61850fe8496e77d075d), so I'm closing the PR.

PR #235, opened 2023-05-29
JBR-5600

VladRassokhin:


VladRassokhin:
It's almost backward compatible, but notarize.sh requires new env variables.

PR #241, opened 2023-08-07
JBR-5824 Ensure popup menus are on the correct screen

stachenov:
This is a very old bug, JDK-6415065.

What happens here is that when the position
of a popup menu is calculated, it can expand
above or below, depending on the position
of the parent menu, the item being expanded
and the size of the submenu and screen resolution.

If the menu decides to expand above,
the position calculation in JMenu.getPopupMenuOrigin may yield a coordinate above the current screen.
Later, JPopupMenu.adjustPopupLocationToFitScreen
tries to fit the entire menu into the screen.
However, it has no idea which screen is correct,
as all it has is an (x, y) location. If that
location is invalid, it may correct it by
fitting it into the screen. However, if it is
valid, but located on an incorrect screen,
then the whole logic goes awry and the menu
is fitted into the wrong screen.

Fix by pre-adjusting the Y location to fit
into the correct screen in JMenu.getPopupMenuOrigin, where the correct screen is still known.
The resulting location may still not be final,
as the menu's height needs to be taken into
account as well, but that's exactly what
JPopupMenu.adjustPopupLocationToFitScreen does.
Since the coordinate is on the correct screen now, it fits the menu into the same screen, which
guarantees it'll be the correct one.

### DIFF
--- a/src/java.desktop/share/classes/javax/swing/JMenu.java
+++ b/src/java.desktop/share/classes/javax/swing/JMenu.java
@@ -478,6 +478,16 @@ public class JMenu extends JMenuItem implements Accessible,MenuElement
                 y = 0 - yOffset - pmSize.height;   // Otherwise drop 'up'
             }
         }
+        // Note that the y position may be later adjusted to fit the menu into the screen if possible.
+        // However, the code that does it (JPopupMenu.adjustPopupLocationToFitScreen) has no idea which screen
+        // to fit into, and determines it by the position calculated here, so we need to make sure it's on
+        // the correct screen, otherwise the menu may appear on the wrong screen (JDK-6415065).
+        if (position.y + y < screenBounds.y) { // Above the current screen?
+            y = screenBounds.y - position.y; // Fit into the screen, relative to our origin.
+        }
+        if (position.y + y >= screenBounds.y + screenBounds.height) { // Below the current screen?
+            y = screenBounds.y + screenBounds.height - 1 - position.y; // Fit into the screen, relative to our origin.
+        }
         return new Point(x,y);
     }
 


src/java.desktop/share/classes/javax/swing/JMenu.java
```diff
@@ -440,6 +440,15 @@ protected Point getPopupMenuOrigin() {
...
+                y = screenBounds.y - position.y; // Fit into the screen, relative to our origin.
+            }
+            if (position.y + y >= screenBounds.y + screenBounds.height) { // Below the current screen?
+                y = screenBounds.y + screenBounds.height - 1 - position.y; // Fit into the screen, relative to our origin.
+            }
+            // Note that the y position will later be adjusted to fit the menu into the screen if possible.
+            // However, the code that does it (JPopupMenu.adjustPopupLocationToFitScreen) has no idea which screen
+            // to fit into, so we at least make sure above that it's on the correct screen already.
         } else {
             // We are a toplevel menu (pull-down)
```
mkartashev, line 453: Is there the same problem with the pull-down menu appearing on another (bottom) screen?

mkartashev (changes requested):

src/java.desktop/share/classes/javax/swing/JMenu.java
```diff
@@ -440,6 +440,15 @@ protected Point getPopupMenuOrigin() {
 
                 y = s.height - yOffset - pmSize.height;
             }
+            if (position.y + y < screenBounds.y) { // Above the current screen?
+                y = screenBounds.y - position.y; // Fit into the screen, relative to our origin.
```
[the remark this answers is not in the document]
stachenov: The coordinates are in `JMenu` coordinate system here, and `position` is assigned `getLocationOnScreen`, so I first convert it to the screen coordinate space here to check (`position.y + y`) and then back. So if I need to set the `y` coordinate to the top edge of the screen, I need `position.y + y` (`y` converted to the screen coordinate space) to be equal to `screenBounds.y`, which means `y = screenBounds.y - position.y`. For example, the screen starts at `screenBounds.y = 1080` and the `JMenu` is at `position.y = 1100`. If `y = -100` this means it's `1000` in the screen coordinate space, which is 80px above the top edge. In this example, I set it to `1080 - 1100 = -20`, which is exactly 20px above the `JMenu`, so in screen coordinates at `1100 - 20 = 1080`. I've actually tested it before pushing too and it reports correct values in the debugger. Maybe it could be rewritten in a better way so we don't have to reason about it, but it'd require refactoring this code, which I'm not sure is a good idea.

mkartashev:
Also, please consider opening a PR against OpenJDK so we don't have to carry this commit with JBR forever.

stachenov:
Will do, but are there any specific workflow requirements for that? For example, do I have to submit a PR to them with this exact commit message, or is it enough if it has exactly the same changes, or what? I just don't know how the merges between OpenJDK and JBR happen, so how to ensure that there are no conflicts when my PR to OpenJDK eventually lands into JBR.

src/java.desktop/share/classes/javax/swing/JMenu.java
mkartashev: Actually, I meant the pull-down menus that are serviced by another branch of the if, lower than the changes you've made

stachenov: Ah, you mean not a popup menu, but a pull-down menu. Hmm... Yes, in theory it could happen. I think the best way would be to move the correcting code out of that `if`. The resulting coordinate must be located on the same screen as the menu in any case, so it definitely won't hurt.

stachenov: Force-pushed the fix that takes both cases into account, just in case the OpenJDK PR takes too long.

mkartashev: Thank you! Looks good to me.

mkartashev (approved):

PR #245, opened 2023-09-21
Fix macOS services writing text back to textfield

ghostflyby:
https://en.wikipedia.org/wiki/Services_menu
Fix macOS menu services not being able to write text back to textfield due to the mismatch of ObjectiveC method signature.

`@selector(setAccessibilitySelectedText)` never selects the desired method `- (void)setAccessibilitySelectedText:(NSString *)accessibilitySelectedText` for lack of the parameter mark (a colon).


tsarn:
Hi, could you please replay your changes over either `jbr17` or `jbr21` branch and make this PR target that branch instead of `main`? It's just that `main` is regularly force-pushed with upstream OpenJDK commits, making it very difficult to see the changes in this PR. Thanks!

ghostflyby:
replaced by #247

PR #247, opened 2023-09-22
JBR-6124 Fix macOS services writing text back to textfield

ghostflyby:
in place of #245 

https://en.wikipedia.org/wiki/Services_menu
Fix macOS menu services not being able to write text back to textfield due to the mismatch of ObjectiveC method signature.

`@selector(setAccessibilitySelectedText)` never selects the desired method `- (void)setAccessibilitySelectedText:(NSString *)accessibilitySelectedText` for lack of the parameter mark (a colon).


mkartashev:
Please, also replace the commit itself with a new one with the updated commit message. Thanks!

PR #249, opened 2023-10-11
JBR-6194 Fix VoiceOver reading old JComboBox value after changing it

dmitrii-drobotov:
### Changes

* Removed `(value == nil)` check in `ComboBoxAccessibility`. With this check, the value was not reassigned on every get request of `accessibilityValue`, but only on get `accessibilitySelectedChildren`. When changing focus by Tab, only get `accessibilityValue` is called, and because `value` is already not nil, the old value was returned. 
    * Even with this check removed, VoiceOver still reads the old value when moving VO cursor from and to the combo box (for example using VO+Left and VO+Right shortcuts). I couldn't find the reason why it happens, but it's possible that when working with a "combo box", VoiceOver expects that there is a text field, and listens to text update events, but in case of non-editable combo box, there is no text field.
* Changed role to `NSAccessibilityPopUpButtonRole` if combo box is not editable. This fixes the bug with not updated value when using VO cursor navigation. Additional reasoning for using "popup button" role:
    * Native MacOS non-editable combo boxes and non-editable HTML `<select>` elements also have the "popup button" role instead of "combo box", so the role should become more clear.
    * With "combo box" role, VoiceOver was saying "Type text, or to display a list of choices, press Control-Option-Space". In case of non-editable combo box, there is no text field, so "type text" instruction was misleading. Now it says only "To display a list of options, press Control-Option-Space".
    * With "popup button" role, shortcut VO+Space to open the combo box menu works correctly.
    
    
### Testing

Tested the fix in SwingSet demo app and in a separate test app (based on AccessibleJComboboxTest but with VoiceOver-specific instructions).

#### Before

SwingSet:

https://github.com/JetBrains/JetBrainsRuntime/assets/102954094/cb59b6f0-4486-4439-a486-04470bf613d8

Test case app:

https://github.com/JetBrains/JetBrainsRuntime/assets/102954094/e606b026-0f08-4b5e-88ee-9b4ea0729e5b

#### After

SwingSet:

https://github.com/JetBrains/JetBrainsRuntime/assets/102954094/e224afd4-c63b-4e2a-a4e7-7b579afdc671

Test case app:

https://github.com/JetBrains/JetBrainsRuntime/assets/102954094/3440492a-9b80-4c6d-9b06-bcdaaf06e054

Additionally tested in IntelliJ with JBR including this fix, the issue [IDEA-317420](https://youtrack.jetbrains.com/issue/IDEA-317420/Accessibility-Voice-Over-on-macOS-reads-old-FontComboBox-value-after-change) is fixed.

### DIFF
--- a/src/java.desktop/macosx/classes/sun/lwawt/macosx/CAccessibility.java
+++ b/src/java.desktop/macosx/classes/sun/lwawt/macosx/CAccessibility.java
@@ -60,6 +60,7 @@ import javax.accessibility.AccessibleTable;
 import javax.accessibility.AccessibleText;
 import javax.accessibility.AccessibleValue;
 import javax.swing.Icon;
+import javax.swing.JComboBox;
 import javax.swing.JComponent;
 import javax.swing.JEditorPane;
 import javax.swing.JLabel;
@@ -1160,6 +1161,20 @@ class CAccessibility implements PropertyChangeListener {
         return invokeAndWait(() -> {
             AccessibleContext ac = a.getAccessibleContext();
             return ac == null ? null : AccessibleComponentAccessor.getAccessible(ac);
+        }, c);
+    }
+
+    private static boolean isComboBoxEditable(Accessible a, Component c) {
+        if (a == null) return false;
+
+        return invokeAndWait(new Callable<Boolean>() {
+            public Boolean call() throws Exception {
+                Accessible sa = CAccessible.getSwingAccessible(a);
+                if (sa instanceof JComboBox<?> comboBox) {
+                    return comboBox.isEditable();
+                }
+                return false;
+            }
         }, c);
     }
 }
--- a/src/java.desktop/macosx/native/libawt_lwawt/awt/a11y/ComboBoxAccessibility.m
+++ b/src/java.desktop/macosx/native/libawt_lwawt/awt/a11y/ComboBoxAccessibility.m
@@ -57,6 +57,18 @@ GET_CACCESSIBILITY_CLASS_RETURN(nil);
     return value;
 }
 
+- (BOOL)isEditable
+{
+    JNIEnv* env = [ThreadUtilities getJNIEnv];
+    GET_CACCESSIBILITY_CLASS_RETURN(NO);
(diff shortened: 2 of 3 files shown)


dmitrii-drobotov:
@NikitkoCent please review.

src/java.desktop/macosx/classes/sun/lwawt/macosx/CAccessibility.java
```diff
@@ -1162,4 +1163,18 @@ private static Accessible getAccessibleJTreeNodeCurrentAccessible(Accessible a,
...
+
+    private static boolean isComboBoxEditable(Accessible a, Component c) {
+        if (a == null) return false;
+
+        return invokeAndWait(new Callable<Boolean>() {
+            public Boolean call() throws Exception {
+                Accessible sa = CAccessible.getSwingAccessible(a);
+                if (sa instanceof JComboBox) {
+                    return ((JComboBox<?>) sa).isEditable();
+                }
```
NikitkoCent, line 1175: Let's use Java 14's enhanced instanceof here 🙏

src/java.desktop/macosx/native/libawt_lwawt/awt/a11y/ComboBoxAccessibility.m
```diff
@@ -66,8 +78,7 @@ - (id)accessibilityValue
     if (expanded) {
         return nil;
     }
-    if (!expanded &&
-        (value == nil)) {
+    if (!expanded) {
         [self accessibleSelection];
     }
```
NikitkoCent, line 83: A few suggestions how we can a bit improve the code here:
* As I see, the check `if (!expanded)` is redundant here, because it has already been checked above.
* Just in case, I'd add a check that the `value` isn't `nil` before invoking any of its methods. It can help us to avoid unexpected JBR crashes.

NikitkoCent (changes requested):
A cool patch! And special thanks for the detailed explanation! Please take a look at the following few minor improvements.

NikitkoCent (approved):

PR #255, opened 2023-11-16
JBR-4479 Add text caret tracking for macOS Accessibility Zoom

dmitrii-drobotov:
Added a call to [UAZoomChangeFocus](https://developer.apple.com/documentation/applicationservices/1458830-uazoomchangefocus?changes=_3&language=objc) function when a text component fires an `ACCESSIBLE_CARET_PROPERTY` changed event. `InputMethodRequests.getTextLocation` is used to get the caret location.

How it works in IDEA and SwingSet (video quality is not great due to GitHub limits):

https://github.com/JetBrains/JetBrainsRuntime/assets/102954094/4590f056-9b00-4b7f-b735-1de51ee9c6f6

https://github.com/JetBrains/JetBrainsRuntime/assets/102954094/4670baab-ae63-4f80-967d-e6dd9ab4426c

### Known issues and further plans
* There are a few cases when zoom doesn't follow caret in the Editor component in IntelliJ: when deleting text, pasting text, adding/removing tab indentation, using page up/down keys, and there may be more. Some of it is caused by the lack of caret change event, which we can implement as follow-up work later, while for other cases, it doesn't seem to work anywhere (e.g. Safari, Chrome, TextEdit).
* Caret tracking works by moving the mouse pointer to the middle of the screen, which can block a part of the text, especially if the pointer size is big. Other apps are actually hiding the pointer when typing or navigating the text, but IntelliJ only hides it when typing (https://youtrack.jetbrains.com/issue/IDEA-69397). We should consider additionally implementing hiding the cursor when moving the caret, not just on typing.
* Ideally, the Zoom should follow all keyboard focus changes, not only the text caret (if it's enabled in Zoom settings), which can be implemented by using the same `UAZoomChangeFocus` function. I decided to work on this part later in a separate PR, as it seems to be a bigger effort, and the text caret tracking alone brings a lot of value for IDE users.

### DIFF
--- a/src/java.desktop/macosx/native/libawt_lwawt/awt/a11y/NavigableTextAccessibility.m
+++ b/src/java.desktop/macosx/native/libawt_lwawt/awt/a11y/NavigableTextAccessibility.m
@@ -306,6 +306,49 @@ static jmethodID sjm_getAccessibleEditableText = NULL;
     return [super accessibilityParent];
 }
 
+
+/*
+ * Class:     sun_lwawt_macosx_CAccessible
+ * Method:    updateZoomCaretFocus
+ * Signature: (I)V
+ */
+JNIEXPORT void JNICALL Java_sun_lwawt_macosx_CAccessible_updateZoomCaretFocus
+        (JNIEnv *env, jclass jklass, jlong element)
+{
+    if (!UAZoomEnabled()) return;
+
+    JNI_COCOA_ENTER(env);
+        [ThreadUtilities performOnMainThread:@selector(postZoomChangeCaretFocus)
+                                          on:(NavigableTextAccessibility *) jlong_to_ptr(element)
+                                  withObject:nil
+                               waitUntilDone:NO];
+    JNI_COCOA_EXIT(env);
+}
+
+- (void)postZoomChangeCaretFocus
+{
+    AWT_ASSERT_APPKIT_THREAD;
+
+    if (![self isEqual:[NSApp accessibilityFocusedUIElement]]) return;
+
+    JNIEnv *env = [ThreadUtilities getJNIEnv];
+
+    GET_CACCESSIBLETEXT_CLASS();
+    DECLARE_STATIC_METHOD(jm_getCaretRectangle, sjc_CAccessibleText, "getCaretRectangle",
+                          "(Ljavax/accessibility/Accessible;Ljava/awt/Component;)[D");
+    jdoubleArray axCaretBounds = (jdoubleArray) (*env)->CallStaticObjectMethod(env, sjc_CAccessibleText,
+                                                                               jm_getCaretRectangle, fAccessible,
+                                                                               fComponent);
+    CHECK_EXCEPTION();
+    if (axCaretBounds == NULL) return;
+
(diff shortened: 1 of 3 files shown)


src/java.desktop/macosx/native/libawt_lwawt/awt/a11y/NavigableTextAccessibility.m
```diff
@@ -306,6 +306,49 @@ - (id)accessibilityParent
...
+    GET_CACCESSIBLETEXT_CLASS();
+    DECLARE_STATIC_METHOD(jm_getCaretRectangle, sjc_CAccessibleText, "getCaretRectangle",
+                          "(Ljavax/accessibility/Accessible;Ljava/awt/Component;)[D");
+    jdoubleArray axCaretBounds = (jdoubleArray) (*env)->CallStaticObjectMethod(env, sjc_CAccessibleText,
+                                                                               jm_getCaretRectangle, fAccessible,
+                                                                               fComponent);
+    CHECK_EXCEPTION();
+    if (axCaretBounds == NULL) return;
+
+    jdouble *values = (*env)->GetDoubleArrayElements(env, axCaretBounds, 0);
```
NikitkoCent, line 345: It'd be a bit more expressive to use `NULL` instead of 0 since the last parameter is a pointer, not an integer value.

src/java.desktop/macosx/native/libawt_lwawt/awt/a11y/NavigableTextAccessibility.m
```diff
@@ -306,6 +306,49 @@ - (id)accessibilityParent
...
+    CGRect caretRect = CGRectMake(values[0], values[1], values[2], values[3]);
+    (*env)->ReleaseDoubleArrayElements(env, axCaretBounds, values, 0);
```
NikitkoCent, line 347: Since you don't change the array's content, it may be better to pass `JNI_ABORT` instead of `0` to the last parameter.

NikitkoCent (changes requested):
A very clear patch, thanks a lot!
There are a few little improvement suggestions. Also, could you please do the following? 
1. Now your work is about JBR-4479. not JBR-4478. See [this](https://youtrack.jetbrains.com/issue/JBR-4478/No-cursor-tracking-while-using-OS-zoom-feature-with-IDE#focus=Comments-27-8405204.0-0). Could you please update the PR description and the commit(s) message(s) according to this change?
2. (If you haven't done this) it'd be great if you take some extra effort and check how well the fix works in the following cases. I'm mostly worried about correctness of positioning.
2.1. When the screen has non-default scaling settings (`System Preferences` -> `Display` -> `Display Settings`).
2.2. When there are several screens.
2.3. When there are several screens with different scaling settings.

I've checked all the cases on my environment, but I think it's worth checking on another one, just in case.

dmitrii-drobotov:
@NikitkoCent I've tested these scenarios, and it worked well because we get a new caret location every time. If any scaling changes happened between caret updates, we're not going reuse the old positing, but get a new one, which should be calculated according to the current scaling.

NikitkoCent (approved):

PR #263, opened 2024-01-03
JBR-6472 Add default value in CAccessibility.isComboBoxEditable to avoid NPE

dmitrii-drobotov:
`CAccessibility.invokeAndWait` can return null if no default value is specified, so in this case we'll get  `java.lang.NullPointerException: Cannot invoke "java.lang.Boolean.booleanValue()" because the return value of "sun.lwawt.macosx.CAccessibility.invokeAndWait(java.util.concurrent.Callable, java.awt.Component)" is null`. 

With the `false` default value `invokeAndWait` will never return null.


NikitkoCent:
Looks good, thanks!

PR #267, opened 2024-01-11
JBR-6543 Vulkan: migrate current code to pure c

avu:


src/java.desktop/share/native/common/java2d/cvulkan/VKBase.c
```diff
@@ -0,0 +1,644 @@
...
+}
+
+
+jboolean VK_Init(jboolean verb, jint requestedDevice) {
+    verbose = verb;
+    requestedDeviceNumber = requestedDevice;
+    return VKGE_graphics_environment() != NULL;
+}
+
+const char* physicalDeviceTypeString(VkPhysicalDeviceType type)
```
mkartashev, line 132: This one isn't exported by the header, so should be `static`, I guess

src/java.desktop/share/native/common/java2d/cvulkan/VKBase.c
```diff
@@ -0,0 +1,644 @@
...
+        if (apiVersion < REQUIRED_VULKAN_VERSION) {
+            J2dRlsTrace3(J2D_TRACE_ERROR, "Vulkan: Unsupported version. Required at least (%d.%d.%d)\n",
+                         VK_API_VERSION_MAJOR(REQUIRED_VULKAN_VERSION),
+                         VK_API_VERSION_MINOR(REQUIRED_VULKAN_VERSION),
+                         VK_API_VERSION_PATCH(REQUIRED_VULKAN_VERSION))
+            vulkanLibClose();
+            return NULL;
+        }
+
+        geInstance = (VKGraphicsEnvironment*)malloc(sizeof(VKGraphicsEnvironment));
```
mkartashev, line 195: Better check for `malloc` failure as otherwise we're going to crash at the next line.

src/java.desktop/share/native/common/java2d/cvulkan/VKBase.c
```diff
@@ -0,0 +1,644 @@
...
+            NULL,0
+        };
+
+        // Get the number of extensions and layers
+        vkEnumerateInstanceExtensionProperties(NULL, &geInstance->extensionsCount, NULL);
+        geInstance->extensions = (VkExtensionProperties*) malloc(sizeof(VkExtensionProperties)*geInstance->extensionsCount);
+        vkEnumerateInstanceExtensionProperties(NULL, &geInstance->extensionsCount, geInstance->extensions);
+
+        vkEnumerateInstanceLayerProperties(&geInstance->layersCount, NULL);
+        geInstance->layers = (VkLayerProperties*) malloc(sizeof(VkLayerProperties)*geInstance->layersCount);
```
mkartashev, line 210: Also a `malloc` check is required here.

src/java.desktop/share/native/common/java2d/cvulkan/VKBase.c
```diff
@@ -0,0 +1,644 @@
...
+        vkEnumeratePhysicalDevices(geInstance->vkInstance, &geInstance->physicalDevicesCount, NULL);
+
+        if (geInstance->physicalDevicesCount == 0) {
+            J2dRlsTrace(J2D_TRACE_ERROR, "Vulkan: Failed to find GPUs with Vulkan support\n")
+            vulkanLibClose();
+            return NULL;
+        } else {
+            J2dRlsTrace1(J2D_TRACE_INFO, "Vulkan: Found %d physical devices:\n", geInstance->physicalDevicesCount)
+        }
+        geInstance->physicalDevices = malloc(sizeof (VkPhysicalDevice)*geInstance->physicalDevicesCount);
```
mkartashev, line 337: Another `malloc` check is requried.

src/java.desktop/share/native/common/java2d/cvulkan/VKBase.c
```diff
@@ -0,0 +1,644 @@
...
+        PFN_vkGetPhysicalDeviceWaylandPresentationSupportKHR vkGetPhysicalDeviceWaylandPresentationSupportKHR =
+                (PFN_vkGetPhysicalDeviceWaylandPresentationSupportKHR) vulkanLibProc(
+                        geInstance->vkInstance, "vkGetPhysicalDeviceWaylandPresentationSupportKHR");
+        if(vkGetPhysicalDeviceWaylandPresentationSupportKHR == NULL) {
+            J2dRlsTrace(J2D_TRACE_ERROR, "vkGetPhysicalDeviceWaylandPresentationSupportKHR is not supported\n");
+            vulkanLibClose();
+            return NULL;
+        }
+#endif
+        geInstance->devices = (VKLogicalDevice *) malloc(geInstance->physicalDevicesCount*sizeof(VKLogicalDevice));
```
mkartashev, line 384: Another `malloc` check is required.

src/java.desktop/share/native/common/java2d/cvulkan/VKBase.c
```diff
@@ -0,0 +1,644 @@
...
+                J2dRlsTrace(J2D_TRACE_INFO, " - hasTimelineSemaphore not supported, skipped \n")
+                continue;
+            }
+            J2dRlsTrace(J2D_TRACE_INFO, "\n")
+
+            uint32_t queueFamilyCount = 0;
+            vkGetPhysicalDeviceQueueFamilyProperties(
+                    geInstance->physicalDevices[i], &queueFamilyCount, NULL);
+
+            VkQueueFamilyProperties *queueFamilies = malloc(sizeof(VkQueueFamilyProperties) * queueFamilyCount);
```
mkartashev, line 422: Another `malloc` check is required.

src/java.desktop/share/native/common/java2d/cvulkan/VKBase.c
```diff
@@ -0,0 +1,644 @@
...
+            }
+            free(queueFamilies);
+            if (queueFamily == -1) {
+                J2dRlsTrace(J2D_TRACE_INFO, "    --------------------- Suitable queue not found, skipped \n")
+                continue;
+            }
+
+            uint32_t layerCount;
+            vkEnumerateDeviceLayerProperties(geInstance->physicalDevices[i], &layerCount, NULL);
+            VkLayerProperties *layers = (VkLayerProperties *) malloc(layerCount * sizeof(VkLayerProperties));
```
mkartashev, line 461: Another `malloc` check is required.

src/java.desktop/share/native/common/java2d/cvulkan/VKBase.c
```diff
@@ -0,0 +1,644 @@
...
+            vkEnumerateDeviceLayerProperties(geInstance->physicalDevices[i], &layerCount, layers);
+            J2dRlsTrace(J2D_TRACE_VERBOSE, "    Supported device layers:\n")
+            for (uint32_t j = 0; j < layerCount; j++) {
+                J2dRlsTrace1(J2D_TRACE_VERBOSE, "        %s\n", (char *) layers[j].layerName)
+            }
+
+            uint32_t extensionCount;
+            vkEnumerateDeviceExtensionProperties(geInstance->physicalDevices[i], NULL, &extensionCount, NULL);
+            VkExtensionProperties *extensions = (VkExtensionProperties *) malloc(
```
mkartashev, line 470: Another `malloc` check is required.

src/java.desktop/share/native/common/java2d/cvulkan/VKBase.c
```diff
@@ -0,0 +1,644 @@
...
+            if (!hasSwapChain) {
+                J2dRlsTrace(J2D_TRACE_INFO,
+                            "    --------------------- Required VK_KHR_SWAPCHAIN_EXTENSION_NAME not found, skipped \n")
+                continue;
+            }
+
+
+            //geInstance->devices->queueFamily = queueFamily;
+            uint32_t deviceEnabledLayersCount = 0;
+            char **deviceEnabledLayers = malloc(MAX_ENABLED_LAYERS * sizeof(char *));
```
mkartashev, line 515: Another `malloc` check is required.

src/java.desktop/share/native/common/java2d/cvulkan/VKBase.c
```diff
@@ -0,0 +1,644 @@
...
+            uint32_t deviceEnabledExtensionsCount = 0;
+            char **deviceEnabledExtensions = malloc(MAX_ENABLED_EXTENSIONS * sizeof(char *));
```
mkartashev, line 517: Another `malloc` check is required.

src/java.desktop/share/native/common/java2d/cvulkan/VKBase.c
```diff
@@ -0,0 +1,644 @@
...
+            geInstance->devices[geInstance->devicesCount].physicalDevice = geInstance->physicalDevices[i];
+            geInstance->devices[geInstance->devicesCount].queueFamily = queueFamily;
+            geInstance->devices[geInstance->devicesCount].enabledLayers = deviceEnabledLayers;
+            geInstance->devices[geInstance->devicesCount].enabledLayersCount = deviceEnabledLayersCount;
+            geInstance->devices[geInstance->devicesCount].enabledExtensions = deviceEnabledExtensions;
+            geInstance->devices[geInstance->devicesCount].enabledExtensionsCount = deviceEnabledExtensionsCount;
+            geInstance->devices[geInstance->devicesCount].hasExtMemoryBudget = hasExtMemoryBudget;
+            geInstance->devices[geInstance->devicesCount].hasKhrSynchronization2 = hasKhrSynchronization2;
+            geInstance->devices[geInstance->devicesCount].hasKhrDynamicRendering = hasKhrDynamicRendering;
+            geInstance->devices[i].name = malloc(strlen(deviceProperties2.properties.deviceName) + 1);
```
mkartashev, line 555: Another `malloc` check is required.

mkartashev: Alternatively, you can use `strdup` instead of malloc/strcpy

src/java.desktop/share/native/common/java2d/cvulkan/VKBase.c
```diff
@@ -0,0 +1,644 @@
...
+            NULL,
+            NULL, 0,
+            NULL, 0, 0,
+            NULL, 0,
+            NULL,0
+        };
+
+        // Get the number of extensions and layers
+        vkEnumerateInstanceExtensionProperties(NULL, &geInstance->extensionsCount, NULL);
+        geInstance->extensions = (VkExtensionProperties*) malloc(sizeof(VkExtensionProperties)*geInstance->extensionsCount);
```
mkartashev, line 206: Another `malloc` check is required.

mkartashev (changes requested):

src/java.desktop/share/native/common/java2d/cvulkan/VKRenderQueue.h
```diff
@@ -0,0 +1,72 @@
...
+#define NEXT_FLOAT(buf)     NEXT_VAL(buf, jfloat)
+#define NEXT_BOOLEAN(buf)   (jboolean)NEXT_INT(buf)
+#define NEXT_LONG(buf)      NEXT_VAL(buf, jlong)
+#define NEXT_DOUBLE(buf)    NEXT_VAL(buf, jdouble)
+#define NEXT_SURFACE(buf) ((VKSDOps*) (SurfaceDataOps*) jlong_to_ptr(NEXT_LONG(buf)))
+
+/*
+ * Increments a pointer (buf) by the given number of bytes.
+ */
+#define SKIP_BYTES(buf, numbytes) (buf) += (numbytes)
```
mkartashev, line 46: Maybe cast buf to char* to be on the safe side and ensure that it is bytes that are skipped?

src/java.desktop/share/native/common/java2d/cvulkan/VKSurfaceData.h
```diff
@@ -0,0 +1,78 @@
...
+
+#ifndef VKSurfaceData_h_Included
+#define VKSurfaceData_h_Included
+
+#include <pthread.h>
+#include "jni.h"
+#include "SurfaceData.h"
+#include "VKBase.h"
+
+typedef struct _VKSDOps VKSDOps;
```
mkartashev, line 35: Strictly speaking, an identifier starting with an underscore followed by an uppercase letter is reserved in C, so it's better to have the underscore at the end.

mkartashev (approved):

YaaZ (changes requested):

src/java.desktop/share/native/common/java2d/cvulkan/VKBase.c
```diff
@@ -0,0 +1,709 @@
...
+        J2dRlsTrace1(J2D_TRACE_ERROR, "%s is not supported\n", procName)
+        return NULL;
+    }
+    return vkProc;
+}
+
+
+jboolean VK_Init(jboolean verb, jint requestedDevice) {
+    verbose = verb;
+    requestedDeviceNumber = requestedDevice;
```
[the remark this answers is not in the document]
avu: I think we need to support both ways, probably we can have special invalid value like -1 to automatically select the best device.

src/java.desktop/share/native/common/java2d/cvulkan/VKBase.c
```diff
@@ -0,0 +1,709 @@
...
+        }
+
+        geInstance = (VKGraphicsEnvironment*)malloc(sizeof(VKGraphicsEnvironment));
+        if (geInstance == NULL) {
+            J2dRlsTrace(J2D_TRACE_ERROR, "Vulkan: Cannot allocate VKGraphicsEnvironment\n")
+            vulkanLibClose();
+            return NULL;
+        }
+
+        *geInstance = (VKGraphicsEnvironment) {
```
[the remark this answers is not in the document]
avu: I'm not sure how calloc can help here, but I can put some comments describing each default value. The reason of the approach above is not to miss initializer after possible extension of the structure

src/java.desktop/share/native/common/java2d/cvulkan/VKSurfaceData.c
```diff
@@ -0,0 +1,44 @@
...
+#include "SurfaceData.h"
+
+/**
+ * These are shorthand names for the surface type constants defined in
+ * VKSurfaceData.java.
+ */
+#define VKSD_UNDEFINED       sun_java2d_pipe_hw_AccelSurface_UNDEFINED
+#define VKSD_WINDOW          sun_java2d_pipe_hw_AccelSurface_WINDOW
+#define VKSD_TEXTURE         sun_java2d_pipe_hw_AccelSurface_TEXTURE
+#define VKSD_RT_TEXTURE      sun_java2d_pipe_hw_AccelSurface_RT_TEXTURE
```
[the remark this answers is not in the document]
avu: These types have simple meaning: window, usual texture, texture that can be render target. I suppose that it could be useful for us to have such abstractions.

src/java.desktop/share/native/common/java2d/cvulkan/VKBase.c
```diff
@@ -755,13 +817,100 @@ jboolean VK_CreateLogicalDevice(jint requestedDevice) {
...
+
+    dependencies[1].srcSubpass = 0;
+    dependencies[1].dstSubpass = VK_SUBPASS_EXTERNAL;
+    dependencies[1].srcStageMask = VK_PIPELINE_STAGE_COLOR_ATTACHMENT_OUTPUT_BIT;
+    dependencies[1].dstStageMask = VK_PIPELINE_STAGE_BOTTOM_OF_PIPE_BIT;
+    dependencies[1].srcAccessMask = VK_ACCESS_COLOR_ATTACHMENT_READ_BIT | VK_ACCESS_COLOR_ATTACHMENT_WRITE_BIT;
+    dependencies[1].dstAccessMask = VK_ACCESS_MEMORY_READ_BIT;
+    dependencies[1].dependencyFlags = VK_DEPENDENCY_BY_REGION_BIT;
+
+    VkRenderPassCreateInfo renderPassInfo = {};
```
YaaZ, line 905: Overall, render passes are too inflexible for our use case, we need to support them as a backup, but prefer dynamic rendering extension when supported.

avu: Let's start with minimal supported API and then provide advanced features.

src/java.desktop/share/native/common/java2d/cvulkan/VKBase.c
```diff
@@ -755,13 +817,100 @@ jboolean VK_CreateLogicalDevice(jint requestedDevice) {
...
-        J2dRlsTrace1(J2D_TRACE_INFO, "Logical device (%s) created\n",
-                     geInstance->devices[geInstance->enabledDeviceNum].name)
     }
+    VKLogicalDevice* logicalDevice = &geInstance->devices[geInstance->enabledDeviceNum];
+    VkDevice device = logicalDevice->device;
+    J2dRlsTrace1(J2D_TRACE_INFO, "Logical device (%s) created\n", logicalDevice->name)
+
+    VkPipelineCacheCreateInfo pipelineCacheCreateInfo = {};
+    pipelineCacheCreateInfo.sType = VK_STRUCTURE_TYPE_PIPELINE_CACHE_CREATE_INFO;
+    if (vkCreatePipelineCache(device, &pipelineCacheCreateInfo, NULL, &logicalDevice->pipelineCache) != VK_SUCCESS)
```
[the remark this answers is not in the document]
avu: Let's keep it here for now

src/java.desktop/share/native/common/java2d/cvulkan/VKBase.c
```diff
@@ -646,71 +700,79 @@ VKGraphicsEnvironment* VKGE_graphics_environment() {
...
+    if(vkCreateDevice == NULL || vkCreatePipelineCache == NULL || vkCreateRenderPass == NULL)
     {
         J2dRlsTrace(J2D_TRACE_ERROR, "Required api is not supported\n")
         vulkanLibClose();
         return JNI_FALSE;
     }
+
+    requestedDeviceNumber = (requestedDeviceNumber == -1) ? 0 : requestedDeviceNumber;
+
+    if (requestedDeviceNumber < 0 || (uint32_t)requestedDeviceNumber >= geInstance->devicesCount) {
```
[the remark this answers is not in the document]
avu: OK, let's keep this way for now and reconsider later if necessary.

src/java.desktop/share/native/common/java2d/cvulkan/VKSurfaceData.c
```diff
@@ -0,0 +1,176 @@
...
+        vkwinsdo->presentModesKhr = calloc(vkwinsdo->presentModeKhrCount, sizeof(VkPresentModeKHR));
+        ge->vkGetPhysicalDeviceSurfacePresentModesKHR(physicalDevice, vkwinsdo->surface, &vkwinsdo->presentModeKhrCount,
+                                                      vkwinsdo->presentModesKhr);
+
+        VkExtent2D extent = {
+                (uint32_t) (vkwinsdo->vksdOps.width),
+                (uint32_t) (vkwinsdo->vksdOps.height)
+        };
+
+        uint32_t imageCount = vkwinsdo->capabilitiesKhr.minImageCount + 1;
```
YaaZ, line 79: What is the reason behind this `minImageCount + 1`? I think we should decide on the number of images in conjunction with the `presentMode`

src/java.desktop/share/native/common/java2d/cvulkan/VKSurfaceData.c
```diff
@@ -0,0 +1,176 @@
...
+        if (vkwinsdo->capabilitiesKhr.maxImageCount > 0 && imageCount > vkwinsdo->capabilitiesKhr.maxImageCount) {
+            imageCount = vkwinsdo->capabilitiesKhr.maxImageCount;
+        }
+        VkSwapchainCreateInfoKHR createInfoKhr = {};
+        createInfoKhr.sType = VK_STRUCTURE_TYPE_SWAPCHAIN_CREATE_INFO_KHR;
+        createInfoKhr.surface = vkwinsdo->surface;
+        createInfoKhr.minImageCount = imageCount;
+        createInfoKhr.imageFormat = vkwinsdo->formatsKhr[0].format;
+        createInfoKhr.imageColorSpace = vkwinsdo->formatsKhr[0].colorSpace;
```
YaaZ, line 88: TODO: the first reported format is not always the easiest "default" format, so we will need some more logic here.
We also need to use this format when creating the render pass, currently we are using hardcoded `VK_FORMAT_B8G8R8A8_UNORM` there

src/java.desktop/share/native/common/java2d/cvulkan/VKSurfaceData.c
```diff
@@ -0,0 +1,176 @@
...
+        createInfoKhr.imageExtent = extent;
+        createInfoKhr.imageArrayLayers = 1;
+        createInfoKhr.imageUsage = VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT;
+        createInfoKhr.imageSharingMode = VK_SHARING_MODE_EXCLUSIVE;
+        createInfoKhr.queueFamilyIndexCount = 0;
+        createInfoKhr.pQueueFamilyIndices = NULL;
+
+        createInfoKhr.preTransform = vkwinsdo->capabilitiesKhr.currentTransform;
+        createInfoKhr.compositeAlpha = VK_COMPOSITE_ALPHA_OPAQUE_BIT_KHR;
+        createInfoKhr.presentMode = VK_PRESENT_MODE_FIFO_KHR;
```
YaaZ, line 98: TODO: inspect available modes and choose something better?

src/java.desktop/share/native/common/java2d/cvulkan/VKSurfaceData.c
```diff
@@ -0,0 +1,176 @@
...
+                                        vkwinsdo->swapChainImages) != VK_SUCCESS) {
+            J2dRlsTrace(J2D_TRACE_ERROR, "Cannot get swapchain images\n");
+            return;
+        }
+        vkwinsdo->swapChainImageFormat = vkwinsdo->formatsKhr[0].format;
+        vkwinsdo->swapChainExtent = extent;
+
+// Create image views
+        vkwinsdo->swapChainImageViewsCount = vkwinsdo->swapChainImagesCount;
+        vkwinsdo->swapChainImageViews = calloc(vkwinsdo->swapChainImageViewsCount, sizeof(VkImageView));
```
YaaZ, line 127: Maybe use new dynamic arrays here?

YaaZ (changes requested):
Overall it's much cleaner now, C API is not THAT bad after all :)

src/java.desktop/share/native/common/java2d/cvulkan/VKSurfaceData.c
YaaZ: Also, `VK_FORMAT_B8G8R8A8_UNORM` may be wrong too. If we want to work in sRGB color space (do we?), then we can get gamma correction done by the hardware with `_SRGB` formats.

src/java.desktop/share/native/common/java2d/cvulkan/CArrayUtil.h
```diff
@@ -0,0 +1,65 @@
+#ifndef C_ARRAY_UTIL_H
+#define C_ARRAY_UTIL_H
+
+#include <stdio.h>
+#include <malloc.h>
+#include <assert.h>
+
+#define ARRAY_CAPACITY_MULT 2
+typedef struct {
+    size_t elem_size;
```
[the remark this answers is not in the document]
avu: It's used for reallocation of the array to provide more space (see CArrayUtil.c CARR_array_realloc)

src/java.desktop/share/native/common/java2d/cvulkan/VKBase.c
```diff
@@ -0,0 +1,965 @@
...
+            .pColorBlendState = &colorBlending,
+            .pDynamicState = &dynamicState,
+            .layout = logicalDevice->pipelineLayout,
+            .renderPass = logicalDevice->renderPass,
+            .subpass = 0,
+            .basePipelineHandle = VK_NULL_HANDLE,
+            .basePipelineIndex = -1
+    };
+
+    if (geInstance->vkCreateGraphicsPipelines(device, VK_NULL_HANDLE, 1, &pipelineInfo, NULL,
```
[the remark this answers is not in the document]
avu: I decided to remove the cache to simplify the code

src/java.desktop/share/native/common/java2d/cvulkan/VKRenderQueue.c
```diff
@@ -0,0 +1,729 @@
...
+            break;
+
+        default:
+            J2dRlsTraceLn1(J2D_TRACE_ERROR,
+                "VKRenderQueue_flushBuffer: invalid opcode=%d", opcode);
+            return;
+        }
+    }
+
+    if (dstOps != NULL && dstOps->drawableType == VKSD_WINDOW) {
```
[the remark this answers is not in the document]
avu: FLUSH_SURFACE is used before the disposal of the surface. We put the rendering here to commit all the changes that were rendered into the window surface buffer

src/java.desktop/share/native/common/java2d/cvulkan/VKSurfaceData.c
avu: I didn't want to allocate too many images

src/java.desktop/share/native/common/java2d/cvulkan/VKBase.c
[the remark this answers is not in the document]
YaaZ: I think we need it anyway, pipeline creation performance would benefit from it even if we will not store it to the disk (and even more if we will)

src/java.desktop/share/native/common/java2d/cvulkan/CArrayUtil.h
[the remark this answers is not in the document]
YaaZ: Sure, it is only used in reallocation, but this is such a compile time constant which is not usually stored explicitly. When adding an element and trying to reallocate an array we could probably get it from the parameter via sizeof

src/java.desktop/share/native/common/java2d/cvulkan/VKRenderQueue.c
```diff
@@ -0,0 +1,729 @@
...
+        VKWinSDOps *winDstOps = (VKWinSDOps *)dstOps;
+        VKGraphicsEnvironment* ge = VKGE_graphics_environment();
+        VKLogicalDevice* logicalDevice = &ge->devices[ge->enabledDeviceNum];
+
+        ge->vkWaitForFences(logicalDevice->device, 1, &logicalDevice->inFlightFence, VK_TRUE, UINT64_MAX);
+        ge->vkResetFences(logicalDevice->device, 1, &logicalDevice->inFlightFence);
+
+        uint32_t imageIndex;
+        ge->vkAcquireNextImageKHR(logicalDevice->device, winDstOps->swapchainKhr, UINT64_MAX,
```
tsarn, line 635: I would add a TODO: handle swapchain lost/not optimal errors

src/java.desktop/share/native/common/java2d/cvulkan/VKSurfaceData.c
YaaZ: I mean, usual logic for choosing number of images is not to use `minImageCount` as a basis, but rather as a lower bound. `presentMode` should be a basis instead, for example in `IMMEDIATE` mode there's no reason to have more than one image, or there is no reason to have `MAILBOX` with two images, as then it will be identical to `FIFO`

src/java.desktop/share/native/common/java2d/cvulkan/VKRenderQueue.c
YaaZ: True, this will fire on window resize

src/java.desktop/share/native/common/java2d/vulkan/VKBase.c
```diff
@@ -0,0 +1,811 @@
...
+        J2dRlsTrace1(J2D_TRACE_ERROR, "Cannot create device:\n    %s\n",
+                     geInstance->devices[geInstance->enabledDeviceNum].name)
+        vulkanLibClose();
+        return JNI_FALSE;
+    }
+    VkDevice device = logicalDevice->device;
+    J2dRlsTrace1(J2D_TRACE_INFO, "Logical device (%s) created\n", logicalDevice->name)
+
+
+    // Create command pull
```
YaaZ, line 746: Typo "pull" -> "pool". Overall, we would need to refactor this to manage command buffers and synchronization primitives, as having one per device would kill the throughput.

src/java.desktop/share/native/common/java2d/vulkan/VKBase.c
```diff
@@ -0,0 +1,811 @@
...
+    ARRAY_PUSH_BACK(&vertices, ((VKTxVertex){-1.0f, -1.0f, 0.0f, 0.0f}));
+    ARRAY_PUSH_BACK(&vertices, ((VKTxVertex){1.0f, -1.0f, 1.0f, 0.0f}));
+    ARRAY_PUSH_BACK(&vertices, ((VKTxVertex){-1.0f, 1.0f, 0.0f, 1.0f}));
+    ARRAY_PUSH_BACK(&vertices, ((VKTxVertex){1.0f, 1.0f, 1.0f, 1.0f}));
+    logicalDevice->blitVertexBuffer = ARRAY_TO_VERTEX_BUF(vertices);
+    if (!logicalDevice->blitVertexBuffer) {
+        J2dRlsTrace(J2D_TRACE_ERROR, "Cannot create vertex buffer\n")
+        return JNI_FALSE;
+    }
+    ARRAY_FREE(vertices);
```
YaaZ, line 804: Why do we need this? We can render a full-screen quad with a simple vertex shader and without any vertex input at all. That would simplify the pipeline creation and we wouldn't need to create and bind the buffer.

src/java.desktop/share/native/common/java2d/vulkan/VKSurfaceData.c
```diff
@@ -0,0 +1,141 @@
...
+                                  VK_IMAGE_USAGE_TRANSFER_DST_BIT | VK_IMAGE_USAGE_SAMPLED_BIT,
+                                  VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT);
+    if (!vksdo->image)
+    {
+        J2dRlsTrace(J2D_TRACE_ERROR, "Cannot create image\n");
+        return;
+    }
+    
+    VKCVertex* cVertices = ARRAY_ALLOC(VKCVertex, 4);
+    ARRAY_PUSH_BACK(&cVertices, ((VKCVertex){-1.0f, -1.0f, RGBA_TO_L4(vksdo->bgColor)}));
```
YaaZ, line 51: I think we better make color a per-instance attribute instead of per-vertex

src/java.desktop/share/native/common/java2d/vulkan/VKSurfaceData.c
```diff
@@ -0,0 +1,141 @@
...
+    ARRAY_PUSH_BACK(&cVertices, ((VKCVertex){1.0f, -1.0f, RGBA_TO_L4(vksdo->bgColor)}));
+    ARRAY_PUSH_BACK(&cVertices, ((VKCVertex){-1.0f, 1.0f, RGBA_TO_L4(vksdo->bgColor)}));
+    ARRAY_PUSH_BACK(&cVertices, ((VKCVertex){1.0f, 1.0f, RGBA_TO_L4(vksdo->bgColor)}));
+
+    vksdo->fillVertexBuffer = ARRAY_TO_VERTEX_BUF(cVertices);
+    if (!vksdo->fillVertexBuffer) {
+        J2dRlsTrace(J2D_TRACE_ERROR, "Cannot create vertex buffer\n")
+        return;
+    }
```
YaaZ, line 60: Why do we create a full-screen quad vertex buffer for each surface?

src/java.desktop/share/native/common/java2d/vulkan/VKSurfaceData.c
```diff
@@ -0,0 +1,141 @@
...
+
+        if (!vkwinsdo->swapChainImages) {
+          J2dRlsTraceLn(J2D_TRACE_ERROR, "Cannot get swapchain images");
+          return;
+        }
+
+        if (!VKImage_CreateFramebuffer(vkwinsdo->vksdOps.image, logicalDevice->fillTexturePoly->renderPass)) {
+            J2dRlsTraceLn(J2D_TRACE_ERROR, "Cannot create framebuffer for window surface")
+            return;
+        }
```
YaaZ, line 137: I think this belongs in `VKSD_InitImageSurface`

YaaZ (approved):

src/java.desktop/share/native/common/java2d/vulkan/VKBase.c
avu: Yes, we can add a separate shader with embedded coordinates. Here I've reused the shader that will be used for blitting arbitrary areas

YaaZ: Even if we are blitting arbitrary area, if we are doing one draw call per rectangle, it will be much easier to go with the same vertex shader and specify position/size via push constants which will be recorded directly into the command buffer. If we are going to do batch blit with a single draw call, then sure we would better go with ordinary vertex buffer, but then we would need to support growing vertex buffer, maintain a pool or something

src/java.desktop/share/native/common/java2d/vulkan/VKSurfaceData.c
YaaZ: Note: this relates to the previous question about `flat` modifier. If we do need to support gradient with the same pipeline, then we cannot make it a per-instance attribute, of course

src/java.desktop/share/native/common/java2d/vulkan/VKBase.c
avu: I suppose we'll have both. For scan-converted shapes there will be bulk rendering, for rectangular blits there will be one draw call.

src/java.desktop/share/native/common/java2d/vulkan/VKSurfaceData.c
avu: Usually, we have a separate shader for gradients, because they may have multiple stops or they may be radial with multiple stops

src/java.desktop/share/native/common/java2d/vulkan/VKSurfaceData.c
avu: I've not decided at that moment where to put such global data but in case of separate shader we don't need it anyway

src/java.desktop/share/native/common/java2d/cvulkan/VKVertex.h
```diff
@@ -28,15 +28,37 @@
 #define VKVertex_h_Included
 
 #include <vulkan/vulkan.h>
+#include "VKBuffer.h"
+
+#define RGBA_TO_L4(c)              \
```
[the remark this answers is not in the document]
avu: I'll check but looks like we have this problem for metal and opengl.

mkurz:
How ready is Vulkan in the JBR now? Usable?

avu:
@mkurz Not yet it's in the early stage hopefully we'll have something relatively feature-complete in several months.

PR #272, opened 2024-01-23
JBR-6593 Fix UI freezes with JAWS announcements

dmitrii-drobotov:
* Execute `AccessibleAnnouncer.nativeAnnounce` on a background thread on Windows to fix UI freezes. IntelliJ calls this method from EDT, but it doesn't need to run on EDT because on Windows it simply calls screen readers API without interacting with UI components. Additionally, when using a background thread, the JAWS `SayString` method, which previously could have been running for multiple seconds, is now executed immediately as expected, but the root cause of previous delays is unclear.
* In `JawsAnnouncer`, initialize COM library with the multithreaded model to allow executing it from different threads. By making `comInitializer` non-static, now COM is initialized and uninitialized on every call of the method as required by the [documentation](https://learn.microsoft.com/en-us/windows/win32/learnwin32/initializing-the-com-library): "Each thread that uses a COM interface must make a separate call to this function. For every successful call to CoInitializeEx, you must call CoUninitialize before the thread exits". The `IJawsApi` COM object is still static and reused by different threads, which is allowed with a multithreaded concurrency model. It shouldn't cause issues because it has no state and only forwards calls to JAWS.

Tested on jbr17, jbr21 and main.

### DIFF
--- a/src/java.desktop/windows/native/libawt/windows/AccessibleAnnouncer/JawsAnnouncer.cpp
+++ b/src/java.desktop/windows/native/libawt/windows/AccessibleAnnouncer/JawsAnnouncer.cpp
@@ -31,7 +31,6 @@
 #include "sun_swing_AccessibleAnnouncer.h"
 #include "jni_util.h"                           // JNU_ThrowOutOfMemoryError
 #include "debug_assert.h"                       // DASSERT
-#include <windows.h>                            // GetCurrentThreadId
 #include <initguid.h>                           // DEFINE_GUID
 
 
@@ -57,7 +56,7 @@ public: // assignments
 public:
     HRESULT tryInitialize() {
         if (!isInitialized()) {
-            m_initializeResult = CoInitialize(nullptr);
+            m_initializeResult = CoInitializeEx(nullptr, COINIT_MULTITHREADED);
         }
         return m_initializeResult;
     }
@@ -109,17 +108,7 @@ bool JawsAnnounce(JNIEnv *env, jstring str, jint priority)
     DASSERT(env != nullptr);
     DASSERT(str != nullptr);
 
-    static const DWORD comInitThreadId = ::GetCurrentThreadId();
-
-    const DWORD currThread = ::GetCurrentThreadId();
-    if (currThread != comInitThreadId) {
-#ifdef DEBUG
-        fprintf(stderr, "JawsAnnounce: currThread != comInitThreadId.\n");
-#endif
-        return false;
-    }
-
-    static ComInitializationWrapper comInitializer;
+    ComInitializationWrapper comInitializer;
     comInitializer.tryInitialize();
     if (!comInitializer.isInitialized()) {
 #ifdef DEBUG
--- a/src/java.desktop/share/classes/sun/swing/AccessibleAnnouncer.java
+++ b/src/java.desktop/share/classes/sun/swing/AccessibleAnnouncer.java
@@ -30,6 +30,8 @@ import sun.awt.AWTThreading;
 
 import javax.accessibility.Accessible;
 import java.lang.annotation.Native;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 
 /**
   * This class provides the ability to speak a given string using screen readers.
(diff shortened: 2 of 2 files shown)


src/java.desktop/windows/native/libawt/windows/AccessibleAnnouncer/JawsAnnouncer.cpp
```diff
@@ -109,17 +108,7 @@ bool JawsAnnounce(JNIEnv *env, jstring str, jint priority)
...
-    const DWORD currThread = ::GetCurrentThreadId();
-    if (currThread != comInitThreadId) {
-#ifdef DEBUG
-        fprintf(stderr, "JawsAnnounce: currThread != comInitThreadId.\n");
-#endif
-        return false;
-    }
-
-    static ComInitializationWrapper comInitializer;
+    ComInitializationWrapper comInitializer;
```
NikitkoCent, line 111: We can try to use TLS to avoid initializing/uninitializing COM on each call of the announcing API:
```cpp
static THREAD_LOCAL ComInitializationWrapper comInitializer;
```
AFAIU, the JAWS announcing API is gonna be used on one specific thread only (which isn't EDT, but still only one specific), hence such an improvement will keep only one thread with initialized COM through the lifetime of the app.

NikitkoCent (changes requested):
There're one important question and one not important improvement suggestion.

src/java.desktop/windows/native/libawt/windows/AccessibleAnnouncer/JawsAnnouncer.cpp
```diff
@@ -57,7 +56,7 @@ class ComInitializationWrapper final {
 public:
     HRESULT tryInitialize() {
         if (!isInitialized()) {
-            m_initializeResult = CoInitialize(nullptr);
+            m_initializeResult = CoInitializeEx(nullptr, COINIT_MULTITHREADED);
```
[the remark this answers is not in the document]
dmitrii-drobotov: I tested it now by changing the boot java runtime in a production build and it works fine, `CoInitializeEx` still returns `S_OK`. The jump list feature, which I assume uses `winshellintegrationbridge`, also works well. 

The [documentation](https://learn.microsoft.com/en-us/windows/win32/api/combaseapi/nf-combaseapi-coinitializeex) doesn't mention this specific case, but it seems that `CoInitializeEx` is separate for every thread, so if we initialize it differently from different threads, it should be fine.

src/java.desktop/windows/native/libawt/windows/AccessibleAnnouncer/JawsAnnouncer.cpp
dmitrii-drobotov: Good idea, I added it. I think it's possible that it will use a different thread later based on https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/concurrent/Executors.html#newSingleThreadExecutor(): "Note however that if this single thread terminates due to a failure during execution prior to shutdown, a new one will take its place if needed to execute subsequent tasks". But because it will a be a single thread at a time, thread_local should work.

NikitkoCent (changes requested):
Asked for the `thread_local` -> `THREAD_LOCAL` change

src/java.desktop/windows/native/libawt/windows/AccessibleAnnouncer/JawsAnnouncer.cpp
dmitrii-drobotov: `THREAD_LOCAL` causes compilation issues, that's why I assumed it should be `thread_local`. Looks like it's defined in hotspot module and not accessible here:
```
src\java.desktop\windows\native\libawt\windows\AccessibleAnnouncer\JawsAnnouncer.cpp(111): error C4430: missing type specifier - int assumed. Note: C++ does not support default-int
src\java.desktop\windows\native\libawt\windows\AccessibleAnnouncer\JawsAnnouncer.cpp(111): error C2146: syntax error: missing ';' before identifier 'ComInitializationWrapper'
```

Should I remove it then?

dmitrii-drobotov: Done.

NikitkoCent (approved):

PR #280, opened 2024-01-31
JBR-6598 Wayland: window gets un-maximized after switching

mkartashev:
When the size of the buffer changes, cancel the frame callback and make sure that the next surface commit happens with the new buffer.


avu (approved):
Looks good

PR #283, opened 2024-02-01
JBR-6617 Wayland: java/awt/Frame/HugeFrame/HugeFrame.java crashes JVM

mkartashev:
Artificially constrain the maximum size of the peer based on the available screen space; Wayland doesn't provide any hints or restrictions for the buffer size.

On the native side, throw OOME early as creating very large buffers may freeze the entire operating system like it happens with XWayland on my machine.

[JBR-6617](https://youtrack.jetbrains.com/issue/JBR-6617) Wayland: java/awt/Frame/HugeFrame/HugeFrame.java crashes JVM


avu (approved):
Looks good!

PR #293, opened 2024-02-07
JBR-6469 Wayland: java/awt/image/ColorModel/DrawCustomColorModel.java throws UnsupportedOperationException

mkartashev:
[JBR-6469](https://youtrack.jetbrains.com/issue/JBR-6469) Wayland: java/awt/image/ColorModel/DrawCustomColorModel.java throws UnsupportedOperationException

When surface's type is `SurfaceType.IntArgb`, an unusual blitting loop called `OpaqueCopyArgbToAny` is used to draw an image onto Frame's graphics. That loop requires the target surface to provide its raster, which no platform surfaces really implement and `WLSMSurfaceData.getRaster()` throws UOE.

With an experiment I proved that at least the GNOME Shell implementation of the Wayland server expects memory buffer to have their alpha pre-multiplied, so the correct surface type should be `SurfaceType.IntArgbPre` anyway and that avoids that loop and makes the test pass. I also corrected the color model to indicate that the color components have their alpha pre-multiplied.

This gives correct color values when a semi-transparent Frame is placed on top of some other window (checked by taking a screenshot and measuring the color values in GIMP).

### DIFF
--- a/src/java.desktop/unix/classes/sun/awt/wl/WLSMGraphicsConfig.java
+++ b/src/java.desktop/unix/classes/sun/awt/wl/WLSMGraphicsConfig.java
@@ -32,7 +32,9 @@ import sun.java2d.wl.WLSMSurfaceData;
 import sun.util.logging.PlatformLogger;
 
 import java.awt.Transparency;
+import java.awt.color.ColorSpace;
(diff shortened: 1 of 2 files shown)


avu (approved):
Looks good

PR #303, opened 2024-02-15
JBR-6246 Do not overwrite CDS archives after created with jlink

mkartashev:
The CDS files (`*.jsa`) are created with the option `--generate-cds-archive` of `jlink`. The `modules` file at their creation time must be the same at run time; so when they are overwritten with the files from `$IMAGES_DIR` where `modules` is different, we get a warning at run time when those CDS files are used.
The fix is to not overwrite `*.jsa` files after they were created.


vprovodin:
The test `test/jdk/jb/build/CDSArchivesTest.sh` currently checks that jsa-files were included into _jbrsdk_-distributions, but after these changes jsa-files will be included into both distributions _jbr_ and _jbrsdk_. The test should be modified to check both distributions.

vprovodin:
Both _windows-x64_ and _windows-aarch64_ do not have jsa-files. I suspect the issue with _windows-aarch64_ is originated from Microsoft - I was going to submit this issue to JBS

mkartashev:
> ...after these changes jsa-files will be included into both distributions _jbr_ and _jbrsdk_. The test should be modified to check both distributions.

I'm not sure this is the case; if they are included into _jbr_ also, I would prefer to remove them rather than to modify the test. As per our earlier agreement, CDS archives only make sense for __jbrsdk__.

mkartashev:
This commit does not add CDS to __jbr__ bundles and does not create CDS files on Windows/aarch64. Please, take a look:
https://github.com/JetBrains/JetBrainsRuntime/pull/303/commits/bc239df27eac50caf5d5d347d34084821e4e5ef6

vprovodin:
IMPLEMENTOR_VERSION is incorrectly defined in _Linux_ so the test wrongly fails

vprovodin (approved):

mkartashev:
Merged

PR #312, opened 2024-02-21
JBR-5989 Wayland: added more tests to jdk_awt_wayland

mkartashev:


vprovodin:
Should the group `jdk_wayland` that would consist of `jdk_awt_wayland` + some other tests be created as well?
The reasons may be:
1. probably we might have non `awt` tests that need to be executed in regular runs against Wayland builds
2. even when `jdk_wayland` would be the same as `jdk_awt_wayland` it would help in generating run configurations without any workarounds. E.g. in order to generate configurations for `wayland-fastdebug` we have got errors about duplicated configurations related to `jdk_awt_wayland` that requires adding some crutches...

mkartashev:
> Should the group jdk_wayland that would consist of jdk_awt_wayland + some other tests be created as well?

Yes, makes sense. I was going to add some swing tests later also.

vprovodin (approved):

PR #314, opened 2024-02-23
JBR-6746 Fix linux build failing on: "error: 'this' pointer is null"

skybber:


src/hotspot/share/prims/jvmtiEnhancedRedefineClasses.cpp
```diff
@@ -1533,8 +1533,8 @@ void VM_EnhancedRedefineClasses::calculate_instance_update_information(Klass* ne
         const InjectedField* const injected = JavaClasses::get_injected(old_klass->name(), &num_injected);
         for (int i = java_fields_count; i < java_fields_count+num_injected; i++) {
           FieldInfo *old_field = old_klass->field(i);
-          if (old_field->is_internal() && old_field->name(NULL) == internal_field->name(NULL)) {
-            copy(old_field->offset(), type2aelembytes(Signature::basic_type(internal_field->signature(NULL))));
+          if (old_field->is_internal() && old_field->name_internal_dcevm() == internal_field->name_internal_dcevm()) {
```
NikitkoCent, line 1536: Just to make sure: there's no need here to additionally check `internal_field->is_internal()`, because it's always supposed to be so, right?

NikitkoCent (changes requested):
Without this patch, `jbr17` branch fails to build on my env (with the errors described in [JBR-6746](https://youtrack.jetbrains.com/issue/JBR-6746/)) with the following configure arguments:
`--enable-option-checking=fatal --with-jtreg=<...> --with-boot-jdk=<...> --with-debug-level=fastdebug --with-conf-name=jbr17-fastdebug`

With the patch applied, jbr17 branch successfully builds with the same configure parameters.

However, there's 1 follow-up question.

src/hotspot/share/prims/jvmtiEnhancedRedefineClasses.cpp
skybber: Yes, there is no need, parameters are passed from `InstanceKlass::do_nonstatic_fields_dcevm` and `internal_field` is always internal field, but your suggestion clarifies the code, so I'll incorporate it.

NikitkoCent (approved):

NikitkoCent:
Have already been merged.

PR #323, opened 2024-02-27
JBR-6763 Wayland: handle popup_done event

xdom:
Certain Wayland compositors (wlroots) invalidate xdg_surface after window with popup loses focus. Subsequent attempts to commit the popup window fail with protocol error
"xdg_surface has never been configured".

Handle popup_done event by hiding the popup window and emitting WINDOW_CLOSING Window event.

### DIFF
--- a/src/java.desktop/unix/classes/sun/awt/wl/WLComponentPeer.java
+++ b/src/java.desktop/unix/classes/sun/awt/wl/WLComponentPeer.java
@@ -1185,10 +1185,9 @@ public class WLComponentPeer implements ComponentPeer {
     }
 
     void notifyPopupDone() {
-        if (targetIsWlPopup()) {
-            setVisible(false);
-            WLToolkit.postEvent(new WindowEvent((Window) target, WindowEvent.WINDOW_CLOSING));
-        }
+        assert(targetIsWlPopup());
+        setVisible(false);
+        WLToolkit.postEvent(new WindowEvent((Window) target, WindowEvent.WINDOW_CLOSING));
     }
 
     private WLGraphicsDevice getGraphicsDevice() {
--- a/src/java.desktop/unix/classes/sun/awt/wl/WLComponentPeer.java
+++ b/src/java.desktop/unix/classes/sun/awt/wl/WLComponentPeer.java
@@ -1184,6 +1184,13 @@ public class WLComponentPeer implements ComponentPeer {
         checkIfOnNewScreen();
     }
 
+    void notifyPopupDone() {
+        if (targetIsWlPopup()) {
+            setVisible(false);
+            WLToolkit.postEvent(new WindowEvent((Window) target, WindowEvent.WINDOW_CLOSING));
+        }
+    }
+
     private WLGraphicsDevice getGraphicsDevice() {
         int scale = 0;
         WLGraphicsDevice theDevice = null;
--- a/src/java.desktop/unix/classes/sun/awt/wl/WLComponentPeer.java
+++ b/src/java.desktop/unix/classes/sun/awt/wl/WLComponentPeer.java
@@ -165,7 +165,9 @@ public class WLComponentPeer implements ComponentPeer {
     }
 
     boolean isVisible() {
-        return visible && hasSurface();
+        synchronized (getStateLock()) {
+            return visible && hasSurface();
+        }
     }
 
     boolean hasSurface() {
@@ -236,7 +238,9 @@ public class WLComponentPeer implements ComponentPeer {
     }
 
     protected void wlSetVisible(boolean v) {
-        this.visible = v;
+        synchronized (getStateLock()) {
+            this.visible = v;
+        }
         if (v) {
             String title = getTitle();
             boolean isWlPopup = targetIsWlPopup();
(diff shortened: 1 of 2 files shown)


mkartashev:
Thank you for reporting the problem and the solution! I'm going to need some time to set up wlroots and then I will get back to you

mkartashev:
I have not been able to reproduce that on the version of Sway that I have: 1.7. Can provide some more details on how to reproduce the issue?

src/java.desktop/unix/classes/sun/awt/wl/WLComponentPeer.java
```diff
@@ -1184,6 +1184,13 @@ void notifyLeftOutput(int wlOutputID) {
         checkIfOnNewScreen();
     }
 
+    void notifyPopupDone() {
+        if (targetIsWlPopup()) {
+            setVisible(false);
+            WLToolkit.postEvent(new WindowEvent((Window) target, WindowEvent.WINDOW_CLOSING));
```
mkartashev, line 1190: Normally, popups do not generate this event when closing. Are you certain this is needed?

src/java.desktop/unix/classes/sun/awt/wl/WLComponentPeer.java
```diff
@@ -1184,6 +1184,13 @@ void notifyLeftOutput(int wlOutputID) {
         checkIfOnNewScreen();
     }
 
+    void notifyPopupDone() {
+        if (targetIsWlPopup()) {
+            setVisible(false);
```
mkartashev, line 1189: Since it is not being called through the "norma" chain (`Component.setVisible()`), we now risk getting another `setVisible(false)` call later, which will probably be harmful. Could you please guard against that in `WLComponentPeer.wlSetVisible()`. Also, the `WLComponentPeer.visible` flag should be accessed under `synchronized(getStateLock())`; could you also add that, please?

src/java.desktop/unix/native/libawt_wlawt/WLComponentPeer.c
```diff
@@ -261,6 +262,14 @@ xdg_popup_done(void *data,
                struct xdg_popup *xdg_popup)
 {
     J2dTrace1(J2D_TRACE_INFO, "WLComponentPeer: xdg_popup_done(%p)\n", xdg_popup);
+    struct WLFrame *frame = (struct WLFrame *) data;
```
mkartashev, line 265: The cast to `(struct WLFrame *)` is unnecessary here.

mkartashev (changes requested):

src/java.desktop/unix/classes/sun/awt/wl/WLComponentPeer.java
mkartashev: Then please leave a comment to that effect above the `postEvent` call; we may need to revisit this later and should know how to reproduce the issue that line is supposed to solve.
Thanks!

src/java.desktop/unix/classes/sun/awt/wl/WLComponentPeer.java
```diff
@@ -236,7 +238,9 @@ private static Window getToplevelFor(Component component) {
     }
 
     protected void wlSetVisible(boolean v) {
-        this.visible = v;
+        synchronized (getStateLock()) {
+            this.visible = v;
```
mkartashev, line 242: There still is a possibility of two consecutive `wlSetVisible(false)` calls; we need a check like this here:
```Java
if (this.visible == v) return;
```

mkartashev (changes requested):

src/java.desktop/unix/classes/sun/awt/wl/WLComponentPeer.java
mkartashev:
```suggestion
            if (this.visible == v) return;
            
            this.visible = v;
```

src/java.desktop/unix/classes/sun/awt/wl/WLComponentPeer.java
mkartashev:
```suggestion
            // TODO: may need a better way of notifying interested components about popup disappearance
            WLToolkit.postEvent(new WindowEvent((Window) target, WindowEvent.WINDOW_CLOSING));
```

src/java.desktop/unix/classes/sun/awt/wl/WLComponentPeer.java
```diff
@@ -1184,6 +1188,12 @@ void notifyLeftOutput(int wlOutputID) {
         checkIfOnNewScreen();
     }
 
+    void notifyPopupDone() {
+        assert(targetIsWlPopup());
+        setVisible(false);
+        WLToolkit.postEvent(new WindowEvent((Window) target, WindowEvent.WINDOW_CLOSING));
```
mkartashev, line 1194:
```suggestion
        // TODO: may need a better way of notifying interested components about popup disappearance
        WLToolkit.postEvent(new WindowEvent((Window) target, WindowEvent.WINDOW_CLOSING));
```

mkartashev:
@xdom If you accept the edits I suggested, we can move forward with this PR.

avu (approved):
LGTM

mkartashev (approved):

mkartashev:
@xdom I'm now synchronizing Wayland-related work with the [project Wakefield repository](https://github.com/openjdk/wakefield/). Since this patch was contributed by you, I can't simply cherry-pick it to OpenJDK. Would you be willing to open a PR against branch `jdk21.0.1-wayland` in that repository? That is, assuming you are or can become an OpenJDK contributor?

PR #325, opened 2024-02-29
JBR-6504 Wayland: optional Robot capability to peek at current window's pixels

mkartashev:
[JBR-6504](https://youtrack.jetbrains.com/issue/JBR-6504) Wayland: optional Robot capability to peek at current window's pixels


avu (approved):
Looks good

PR #328, opened 2024-03-06
JBR-6593 Set unique name for the AccessibleAnnouncer thread

dmitrii-drobotov:
Follow-up on #272. Set unique name for the AccessibleAnnouncer thread for easier debugging.
Thread view in IntelliJ showing AccessibleAnnouncer thread name

### DIFF
--- a/src/java.desktop/share/classes/sun/swing/AccessibleAnnouncer.java
+++ b/src/java.desktop/share/classes/sun/swing/AccessibleAnnouncer.java
@@ -26,6 +26,7 @@
 
 package sun.swing;
 
+import jdk.internal.misc.InnocuousThread;
 import sun.awt.AWTThreading;
 
 import javax.accessibility.Accessible;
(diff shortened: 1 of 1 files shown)


src/java.desktop/share/classes/sun/swing/AccessibleAnnouncer.java
```diff
@@ -82,7 +83,16 @@ public static void announce(Accessible a, final String str, final int priority)
         // so execute it on a background thread.
         if (System.getProperty("os.name").startsWith("Windows")) {
             if (executor == null) {
-                executor = Executors.newSingleThreadExecutor();
+                executor = Executors.newSingleThreadExecutor(new ThreadFactory() {
+                    private final ThreadFactory factory = Executors.defaultThreadFactory();
+                    @Override
+                    public Thread newThread(Runnable r) {
```
mkartashev, line 89: It is customary to use `jdk.internal.misc.InnocuousThread` for JDK internal needs. Consider using that instead of a regular thread. See `jdk.internal.ref.CleanerFactory` for an example.

dmitrii-drobotov: Thanks for advice, I changed it to use `InnocuousThread`.

NikitkoCent (approved):
Looks good, thanks!

PR #336, opened 2024-03-11
JBR-6808 Don't create AccessibleJTreeNode for the tree root if it's not visible

dmitrii-drobotov:
This fixes an issue with `AccessibleJTreeNode#getBounds`, which adjusts the node's bounds according to the parent node:
https://github.com/JetBrains/JetBrainsRuntime/blob/ce6f4d023beef597cb670072e093eeb0fb25dfd5/src/java.desktop/share/classes/javax/swing/JTree.java#L5425-L5437

For nodes whose parent is the invisible root, `getBounds` was returning null, and it caused issues with assistive technology like macOS Accessibility Zoom.

This also matches behavior in `AccessibleJTree#getAccessibleChild` which creates accessible node for the root only if it's visible:

https://github.com/JetBrains/JetBrainsRuntime/blob/ce6f4d023beef597cb670072e093eeb0fb25dfd5/src/java.desktop/share/classes/javax/swing/JTree.java#L4541-L4575

### Before

Invisible root has an item in the hierarchy:

Accessibility hierarchy for JTree showing an item for the root, which is not visible in UI

In IntelliJ it caused issues with macOS Accessibility Zoom when selecting top level nodes:

https://github.com/JetBrains/JetBrainsRuntime/assets/102954094/e689c409-19a2-4341-a92d-b7b78e1a03ae

### After

No item in the hierarchy for the invisible root:

Screenshot 2024-03-11 at 12 20 25

Issues with macOS Accessibility Zoom when selecting top level nodes are fixed:

https://github.com/JetBrains/JetBrainsRuntime/assets/102954094/002daa88-6c8a-48d1-ad69-c27e3f67b3fd

### DIFF
--- a/test/jdk/javax/accessibility/8249806/AccessibleJTreeNodeAccessibleParentTest.java
+++ b/test/jdk/javax/accessibility/8249806/AccessibleJTreeNodeAccessibleParentTest.java
@@ -1,0 +1,99 @@
+/*
+ * Copyright (c) 2024, JetBrains s.r.o.. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+/*
+ * @test
+ * @bug 8249806
+ * @summary Tests that JTree's root is not set as accessible parent of top-level tree nodes (direct children of root)
+ *          when the JTree has setRootVisible(false).
+ * @run main AccessibleJTreeNodeAccessibleParentTest
+ */
+
+import java.awt.Robot;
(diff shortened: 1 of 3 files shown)


NikitkoCent (changes requested):
Thanks for the patch! Could you please supplement the patch with an automatic regression test? AFAIU it shouldn't be difficult in this case.

dmitrii-drobotov:
I've added the test, and also ported the change to the copy of `AccessibleJTreeNode` in AccessBridge.java. Turns out it's used instead of the normal `AccessibleJTreeNode` in case when tree is the parent of the node: https://github.com/JetBrains/JetBrainsRuntime/blob/b3cd00363f3e6022b6fb47823c0128c7f5d86798/src/jdk.accessibility/windows/classes/com/sun/java/accessibility/internal/AccessBridge.java#L5469-L5479

It passes `null` as accessible parent argument, so needs to be fixed it the same way. With this additional fix in AccessBridge, the [JDK-8249806](https://bugs.openjdk.org/browse/JDK-8249806) is completely fixed. 

Before fix:
NVDA reads tree levels starting from 2

After:
NVDA reads tree levels starting from 1

test/jdk/javax/accessibility/8249806/AccessibleJTreeNodeAccessibleParentTest.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    private static void doTest() throws Exception {
+        try {
+            SwingUtilities.invokeAndWait(() -> createGUI());
+            Robot robot = new Robot();
+            robot.waitForIdle();
+
+            SwingUtilities.invokeAndWait(() -> {
+                jTree.getAccessibleContext().addPropertyChangeListener(evt -> {
```
NikitkoCent, line 64: As far as I understand, if the control flow doesn't reach the check `if (!jTree.equals(accessibleNode.getAccessibleContext().getAccessibleParent()))`, it will be also the wrong behavior, although there's no check for this here.

Could you please add such a check? 🙏

NikitkoCent (changes requested):
I've tested the patch on macOS: the changes and the test look stable (reverting the changes in `src` makes the test fail immediately ; applying those changes back makes the test pass stable).

However, there's a request for a little improvement of the test.

test/jdk/javax/accessibility/8249806/AccessibleJTreeNodeAccessibleParentTest.java
dmitrii-drobotov: Good point, I've added it.

NikitkoCent (approved):
Great! LGTM

PR #338, opened 2024-03-13
JBR-6814 Wayland: support sun.java2d.uiScale property

mkartashev:


src/java.desktop/unix/classes/sun/awt/wl/WLComponentPeer.java
```diff
@@ -39,7 +39,26 @@
...
+import java.awt.FontMetrics;
+import java.awt.Frame;
+import java.awt.Graphics;
+import java.awt.GraphicsConfiguration;
+import java.awt.Image;
+import java.awt.Point;
+import java.awt.Rectangle;
+import java.awt.SystemColor;
+import java.awt.Toolkit;
+import java.awt.Window;
```
YaaZ, line 61: Why expand import list?

YaaZ:
Also, did you test this with fractional scaling on/off?

mkartashev:
> Also, did you test this with fractional scaling on/off?

Not initially, but now I have. Works OK, same as without fractional scaling on.

src/java.desktop/unix/classes/sun/awt/wl/WLComponentPeer.java
mkartashev: I think OpenJDK wants its code without wildcard imports.

src/java.desktop/unix/classes/sun/awt/wl/WLComponentPeer.java
```diff
@@ -103,6 +122,7 @@ public class WLComponentPeer implements ComponentPeer {
     int width;  // protected by dataLock
     int height; // protected by dataLock
```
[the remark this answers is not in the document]
mkartashev: > Maybe we better work with device space coordinates in peers? 

An interesting idea. I started to draft it, but it turned out that far more usages require user space rather than device space.

YaaZ (approved):

PR #341, opened 2024-03-19
JBR-6826 Include JDK-8324241 in the JBR17/21 builds

mkartashev:
[JBR-6826](https://youtrack.jetbrains.com/issue/JBR-6826) Include JDK-8324241 in the JBR17/21 builds

The original pull request: https://github.com/openjdk/jdk/pull/17509
The original issue: https://bugs.openjdk.org/browse/JDK-8324241

8324241: Always record evol_method deps to avoid excessive method flushing
Reviewed-by: eastigeevich, phh, coleenp, dlong, shade

This commit passes tier1 tests.

### DIFF
--- a/src/hotspot/share/runtime/globals.hpp
+++ b/src/hotspot/share/runtime/globals.hpp
@@ -1980,6 +1980,14 @@ const int ObjectAlignmentInBytes = 8;
           "(default) disables native heap trimming.")                       \
           range(0, UINT_MAX)                                                \
                                                                             \
+  product(bool, ProfileExceptionHandlers, true,                             \
+          "Profile exception handlers")                                     \
+                                                                            \
+  product(bool, AlwaysRecordEvolDependencies, true, EXPERIMENTAL,           \
+                "Unconditionally record nmethod dependencies on class "     \
+                "rewriting/transformation independently of the JVMTI "      \
+                "can_{retransform/redefine}_classes capabilities.")         \
+                                                                            \
   product(bool, AllowEnhancedClassRedefinition, false,                      \
              "Allow enhanced class redefinition beyond swapping method "    \
              "bodies")                                                      \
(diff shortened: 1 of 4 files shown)


src/hotspot/share/runtime/init.cpp
```diff
@@ -115,6 +115,12 @@ void vm_init_globals() {
 jint init_globals() {
   management_init();
   JvmtiExport::initialize_oop_storage();
+#if INCLUDE_JVMTI
+  if (AlwaysRecordEvolDependencies) {
+    JvmtiExport::set_can_hotswap_or_post_breakpoint(true);
```
skybber, line 120: How it is related to this commit purpose?

src/hotspot/share/runtime/globals.hpp
```diff
@@ -1980,6 +1980,14 @@ const int ObjectAlignmentInBytes = 8;
           "(default) disables native heap trimming.")                       \
           range(0, UINT_MAX)                                                \
                                                                             \
+  product(bool, ProfileExceptionHandlers, true,                             \
+          "Profile exception handlers")                                     \
+                                                                            \
+  product(bool, AlwaysRecordEvolDependencies, true, EXPERIMENTAL,           \
```
skybber, line 1986: I would suggest default value `false`

mkartashev: > I would suggest default value `false`

This would defeat the purpose of this change; with `true`, it is supposed to improve IDEA testing performance where it is generally not possible to supply custom VM arguments AFAIK.

src/hotspot/share/runtime/init.cpp
mkartashev: Perhaps, because of this from `src/hotspot/share/jvmci/jvmciCompilerToVM.cpp`:
```
  if (JvmtiExport::can_hotswap_or_post_breakpoint()) {
    // VM needs to track method dependencies
    flags |= 0x0002;
  }
```

Anyway, that's what the original commit does and I'm very reluctant to modify it as I can't anticipate all the consequences of doing so.

skybber (approved):

PR #343, opened 2024-03-20
JBR-6847 Improve locking performance on Windows

mkartashev:
[JBR-6847](https://youtrack.jetbrains.com/issue/JBR-6847) Improve locking performance on Windows

Some info:
* https://learn.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-waitforsingleobject
* https://learn.microsoft.com/en-us/windows/win32/Sync/critical-section-objects


NikitkoCent (approved):
Looks cool, haven't found any possible issues here. Let's merge!

Speaking of whether the critical section API is broken: I've found only one case that could be related only if OpenJDK restored after OOM _kernel_ exception is thrown from `EnterCriticalSection` and tried to keep using the same instance of the critical section. The problem is described in [the article from Raymond Chen](https://devblogs.microsoft.com/oldnewthing/20140911-00/?p=44103) and has already been fixed since Windows XP:

> The contention event is allocated only when contention occurs. (This is what older versions of MSDN meant when they said that the event is “allocated on demand.”) Which leads to a nasty problem: What if contention occurs, but the attempt to create the contention event fails? Originally, the answer was “The kernel raises an out-of-memory exception.”
> 
> Now you’d think that a clever program could catch this exception and try to recover from it, say, by unwinding everything that led up to the exception. Unfortunately, the weakest link in the chain is the critical section object itself: In the original version of the code, the out-of-memory exception was raised while the critical section was in an unstable state. Even if you managed to catch the exception and unwind everything you could, the critical section was itself irretrievably corrupted.
> 
> [...]
> 
> Windows XP solved the problem of the dreaded “out of memory” exception by using a fallback algorithm that could be used if the contention event could not be allocated. The fallback algorithm is not as efficient, but at least it avoids the “out of memory” situation

mkartashev:
THe commit has been merged manually.

PR #346, opened 2024-03-26
JBR-6895 Wayland: cursor changes to resize at edges even when window is maximized

mkartashev:
[JBR-6895](https://youtrack.jetbrains.com/issue/JBR-6895) Wayland: cursor changes to resize at edges even when window is maximized


avu (approved):
LGTM

PR #348, opened 2024-04-01
JBR-6915 add the option -w into mkimages scripts

vprovodin:
The option `-w` was introduced into mkimages scripts. The option add the parameter `--disable-warnings-as-errors` into `configure`.
Note: `configure` is not invoked when mkimages scripts are started with the option `-i`


mkartashev (approved):

avu (approved):
LGTM

PR #351, opened 2024-04-05
JBR-2702 remove workaround after fixing this issue on Apple's side

vprovodin:
It reverts changes made in `src/java.desktop/macosx/native/libawt_lwawt/awt/AWTView.m` within the commit _JBR-2702 Tooltips display through other applications on hover_. So the test introduced by that commit is still active and is included into regular runs.


vprovodin:
The test successfully passes on _macOS12_, _macOS13_, _macOS13_ on both arch _x86-64_ and _aarch64_

PR #352, opened 2024-04-05
JBR-6926 Wayland: fonts are aliased/grainy on first start

mkartashev:
[JBR-6926](https://youtrack.jetbrains.com/issue/JBR-6926) Wayland: fonts are aliased/grainy on first start

The fix is to update volatile images created for offscreen painting of a `Component` whenever that component appears on a monitor with a different scale.


avu (approved):
LGTM

PR #356, opened 2024-04-09
JBR-6588: Add JBR API for inspecting certain properties of KeyEvents

tsarn:
This adds a JBR API called Keyboard, which for now only contains one function that allows the user to inspect additional properties set in the KeyEvent. This allows to get rid of certain heuristics on client side, like the detection of dead keys. Currently these properties are only set on macOS.


tsarn:
Fixed the PR to be against `jbr21`

tsarn:
I will probably close this for now, and wait for JBR API v3

PR #358, opened 2024-04-12
JBR-6984 Fix BoxLayout/NPECheckRequests test

stachenov:
It was initially written and tested on Linux, but it turns out that on other systems validate() can be called in between init() and start() calls, which would break the test even though BoxLayout isn't broken.

Fix by overriding validate() and ignoring exceptions that were intentionally thrown by us. This both fixes the test and ensures that BoxLayout remains broken until we're ready to test it.


vprovodin (commented):
1. 
```
                } catch (RuntimeException e) {
                    if (foo.broken) {
                        // expected, ignore
                    } else {
                        throw e;
                    }
                }
```
2. 
```
        try {
            frame.pack();
        } catch (RuntimeException ignored) {
            // our broken component threw an exception
        }
```
Shouldn't it be checked that `RuntimeException` originated from `BrokeComponent.getPreferredSize`? Whether other causes cannot lead to `RuntimeException` during the above actions...
Even `RuntimeException` originated from `BrokeComponent.getPreferredSize` I would log it like `RuntimeException was thrown by BrokeComponent.getPreferredSize - expected`

BTW in order to avoid including unrelated commits into PR you should rebase PR-branch on top of the latest state of the branch _main_ and make force-push.

stachenov:
> Shouldn't it be checked that `RuntimeException` originated from `BrokeComponent.getPreferredSize`? Whether other causes cannot lead to `RuntimeException` during the above actions... Even `RuntimeException` originated from `BrokeComponent.getPreferredSize` I would log it like `RuntimeException was thrown by BrokeComponent.getPreferredSize - expected`

Fixed by using a custom exception class instead. Added logging as well.

> BTW in order to avoid including unrelated commits into PR you should rebase PR-branch on top of the latest state of the branch _main_ and make force-push.

That's what I did, but apparently the main branch got updated right afterwards. Force-pushed again with the new fix (or should I squash those instead?).

vprovodin (approved):

mkartashev:
Merged into `main`, `jbr21`, `jbr17`.

PR #364, opened 2024-04-19
JBR-7010 Wayland: Swing window resizing is not smooth enough

mkartashev:
[JBR-7010](https://youtrack.jetbrains.com/issue/JBR-7010) Wayland: Swing window resizing is not smooth enough

By re-using memory buffers and allocating them lazily this commit boosts FPS during interactive resize by ~3x.


avu (approved):
LGTM

PR #368, opened 2024-04-23
JBR-7020 Reorder LCD glyph cache freeing and validation

YaaZ:
1. As we started committing the command buffer on glyph cache flush, this invalidates the current encoder. We need to `MTLTR_ValidateGlyphCache` after the flush, not before.
2. There's no reason to maintain separate glyph cache invalidation logic for this singe case (which is a no-op in reality), so just free the cache instead.


avu (approved):
LGTM

PR #372, opened 2024-04-26
JBR-7028 Implement FPS counter on Linux

mkartashev:
[JBR-7028](https://youtrack.jetbrains.com/issue/JBR-7028) Implement FPS counter on Linux

Please review this from the point of view of usability both on the provider and the supplier side. We will add JBRApi for external access in a separate commit.

I suggest that those toolkits that can provide more or less reliable hardware FPS use the same counter name; here, it is `java2d.native.frames`, but any other will do as well.

### DIFF
--- a/src/java.desktop/share/classes/java/awt/Window.java
+++ b/src/java.desktop/share/classes/java/awt/Window.java
@@ -4507,6 +4507,11 @@ public class Window extends Container implements Accessible {
                 return w.getOwnedWindows_NoClientCode();
             }
 
+            public boolean countersEnabled(Window w) {
+                // May want to selectively enable or disable counters per window
+                return counters != null;
+            }
+
             public void bumpCounter(Window w, String counterName) {
                 Objects.requireNonNull(w);
                 Objects.requireNonNull(counterName);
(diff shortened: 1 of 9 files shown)


avu (approved):
LGTM

src/java.desktop/share/classes/java/awt/Window.java
```diff
@@ -4503,13 +4506,99 @@ public void setTrayIconWindow(Window w, boolean isTrayIconWindow) {
...
+            public void bumpCounter(Window w, String counterName) {
+                Objects.requireNonNull(w);
+                Objects.requireNonNull(counterName);
+
+                PerfCounter newCounter;
+                long curTimeNanos = System.nanoTime();
+                synchronized (w.perfCounters) {
+                    newCounter = w.perfCounters.compute(counterName, (k, v) ->
+                            v == null
+                            ? new PerfCounter(curTimeNanos, 1L)
```
bourgesl, line 4519: Where is the PerfCounter class ? Missing import ?

mkartashev: It's further down at line 4597. Simply a two-field record
```
private record PerfCounter(Long updateTimeNanos, Long value) {}
```

dmitriimorskii (approved):

bourgesl (approved):

mkartashev:
Merged into `jbr21` as 9d011d23dfd61f82892a98a072eb88016debcf50
A separate commit will be prepared for `main` as it lacks the Wayland toolkit.

PR #375, opened 2024-05-02
JBR-7058 Wayland: IDE hang on the popup appearance

mkartashev:
Clean up the damage list when resizing a surface.
Additionally, clamp the damaged area before copying to its current actual size in order to safeguard against invalid external input.


avu (approved):
LGTM

PR #376, opened 2024-05-03
JBR-7047 Deadlock on git fetch on Wayland

mkartashev:
[JBR-7047](https://youtrack.jetbrains.com/issue/JBR-7047) Deadlock on git fetch on Wayland


avu (approved):
LGTM

PR #381, opened 2024-05-09
JBR-7128 Use the correct WmSize event type for JFrame moved to another monitor

stachenov:
On Windows, a JFrame may be moved to another monitor while it's maximized by pressing Win+Shift+Left/Right. When this happens, if the monitor has a different scale factor, then a fake WmSize event is sent. This fake event must use the correct type, matching the actual state of the frame, but it wasn't the case, SIZENORMAL was always used.

Fix by introducing a new virtual function that returns the event type corresponding to the actual state and override this function for AwtFrame. Since it's impossible to move a minimized window, only handle two cases: the frame is maximized or not maximized.


mkartashev:
Is there a youtrack issue for that?

stachenov:
What do you mean is there an issue? It's referenced in the title, as it should be: https://youtrack.jetbrains.com/issue/JBR-7128

dmitriimorskii (approved):

PR #386, opened 2024-05-22
JBR-7194: extension-based filters in native file dialogs

trespasserw:
Adding an ability to set simple file-extension-based filters for file dialogs. While being less flexible than `FileDialog#setFilenameFilter`, extension-based filters don't require callbacks to Java code and are supported on all platforms.


trespasserw:
(WIP; the implementation for macOS comes later)

avu:
@trespasserw It would be nice to have a regression test for this feature (something like test/jdk/jb/javax/swing/CustomTitleBar/JFrameNativeControlsMacOSTest.java). It can be created even for incomplete implementation using the platform tag in the jtreg test case.

vprovodin:
Should the ticket IJPL-150176 be moved into the JBR YT project?

trespasserw:
@vprovodin what are those commits?

IJPL-150176 should not be moved, but I will create a companion ticket in JBR later today and edit commit messages.

src/java.desktop/share/classes/com/jetbrains/desktop/JBRFileDialog.java
```diff
@@ -55,9 +42,14 @@ public int getHints() {
         return hints;
     }
 
-    public void setLocalizationStrings(String openButtonText, String selectFolderButtonText) {
+    public void setLocalizationStrings(String openButtonText, String selectFolderButtonText, String allFilesFilterDescription) {
```
YaaZ, line 45: Removing an API method is a breaking change, I think adding more generic `setLocalizationString(key, value)` would be better

src/java.desktop/share/classes/com/jetbrains/desktop/JBRFileDialog.java
```diff
@@ -25,28 +25,15 @@ public static JBRFileDialog get(FileDialog dialog) {
...
-    /**
-     * Text for "Open" button (used when common file dialogs are enabled on
-     * Windows).
-     */
     public String openButtonText;
-
-    /**
-     * Text for "Select Folder" button (used when common file dialogs are
-     * enabled on Windows).
-     */
```
YaaZ, line 48: What is the purpose of removing this javadoc?

src/jetbrains.api/src/com/jetbrains/JBRFileDialog.java
```diff
@@ -37,12 +42,11 @@ static JBRFileDialog get(FileDialog dialog) {
...
     /**
-     * Set JBR-specific file dialog hints:
+     * Set file dialog hints:
      * <ul>
-     *     <li>SELECT_FILES_HINT, SELECT_DIRECTORIES_HINT - Whether to select files, directories or both
-     *     (used when common file dialogs are enabled on Windows, or on macOS),
-     *     both unset bits are treated as a default platform-specific behavior</li>
-     *     <li>CREATE_DIRECTORIES_HINT - Whether to allow creating directories or not (used on macOS)</li>
+     *     <li>SELECT_FILES_HINT, SELECT_DIRECTORIES_HINT - Whether to select files, directories, or both;
+     *     neither bit set is treated as a default platform-specific behavior</li>
```
YaaZ, line 48: I think original wording was more clear and new one is kinda ambiguous:
* "both unset bits are treated as a default platform-specific behavior" -> if neither of `SELECT_FILES_HINT` or `SELECT_DIRECTORIES_HINT` is set, then behavior is platform-specific
* "neither bit set is treated as a default platform-specific behavior" -> No combination of bits will result in platform-specific behavior?

YaaZ (changes requested):

src/java.desktop/share/classes/com/jetbrains/desktop/JBRFileDialog.java
trespasserw: AFAIU, classes in `com.jetbrains.desktop` are part of the corresponding JBR module which the client code is not supposed to use directly. It is proxied by `com.jetbrains.JBRFileDialog` in `jbr-api`, which preserves compatibility.

A key-based approach is an interesting idea; I'll try.

src/java.desktop/share/classes/com/jetbrains/desktop/JBRFileDialog.java
trespasserw: The comments weren't removed, they were moved from a "backend" class to the "API" one.

src/jetbrains.api/src/com/jetbrains/JBRFileDialog.java
trespasserw: Updated.

src/java.desktop/share/classes/com/jetbrains/desktop/JBRFileDialog.java
YaaZ: That's correct, but as you added default implementation for this method in`jbr-api`, means that just running on a newer JBR (without updating jbr-api dependency) would already be incompatible, as there would be no implementation for that API method. I suggest leaving this method as is, we can just mark is as deprecated for the time being

YaaZ:
Pushed to `main` for now

YaaZ:
Pushed this to `main`, `jbr17`, `jbr21`, so closing the review

PR #391, opened 2024-05-25
JBR-7206 Wayland: Stylepad demo flickers when resizing on KDE

mkartashev:
[JBR-7206](https://youtrack.jetbrains.com/issue/JBR-7206) Wayland: Stylepad demo flickers when resizing on KDE

The main idea is to make sure all kinds of sizes (both of the buffer and of the surface) are set in one and only one surface commit.


avu (approved):
LGTM

PR #394, opened 2024-05-30
JBR-7202 wayland: memory leak when resizing windows

mkartashev:
[JBR-7202](https://youtrack.jetbrains.com/issue/JBR-7202) wayland: memory leak when resizing windows

Instead of relying on the property change event, it is enough to signal `IMAGE_INCOMPATIBLE` when this surface manager gets validated against a graphics configuration with a different transform. This will result in "invalid" buffers dropped by the component's repaint manager and new buffers created with no references that prevent garbage collection.


avu (approved):
LGTM

PR #396, opened 2024-05-31
JBR-7237 Fix cyclic dependency of Wayland and Vulkan initialization

YaaZ:
This adds asynchronous Vulkan initialization.
1. `WLGraphicsEnvironment` starts Vulkan initialization in separate thread and proceeds with Wayland initialization
2. Before `vkGetPhysicalDeviceWaylandPresentationSupportKHR` Vulkan init waits for the `wl_display` to become available
3. `VKInstance.isVulkanEnabled()` which is used later by `WLGraphicsDevice` waits for the Vulkan initialization to finish

### DIFF
--- a/src/java.desktop/share/native/common/java2d/vulkan/VKBase.c
+++ b/src/java.desktop/share/native/common/java2d/vulkan/VKBase.c
@@ -44,7 +44,7 @@ static const uint32_t REQUIRED_VULKAN_VERSION = VK_MAKE_API_VERSION(0, 1, 2, 0);
 #define VALIDATION_LAYER_NAME "VK_LAYER_KHRONOS_validation"
 #define COUNT_OF(x) (sizeof(x)/sizeof(x[0]))
 #if defined(VK_USE_PLATFORM_WAYLAND_KHR)
-static struct wl_display *wl_display;
+static struct wl_display *waylandDisplay;
 #endif
 
 static jboolean verbose;
@@ -158,7 +158,7 @@ void* vulkanLibProc(VkInstance vkInstance, char* procName) {
 JNIEXPORT jboolean JNICALL
 Java_sun_java2d_vulkan_VKInstance_initNative(JNIEnv *env, jclass wlge, jlong nativePtr, jboolean verb, jint requestedDevice) {
     #if defined(VK_USE_PLATFORM_WAYLAND_KHR)
-    wl_display = (struct wl_display*) jlong_to_ptr(nativePtr);
+    waylandDisplay = (struct wl_display*) jlong_to_ptr(nativePtr);
     #endif
     verbose = verb;
     if (VKGE_graphics_environment() == NULL) {
@@ -554,7 +554,7 @@ jboolean VK_FindDevices() {
 #if defined(VK_USE_PLATFORM_WAYLAND_KHR)
             VkBool32 presentationSupported =
                 geInstance->vkGetPhysicalDeviceWaylandPresentationSupportKHR(
-                    geInstance->physicalDevices[i], j, wl_display);
+                    geInstance->physicalDevices[i], j, waylandDisplay);
 #endif
             char logFlags[5] = {
                     queueFamilies[j].queueFlags & VK_QUEUE_GRAPHICS_BIT ? 'G' : '-',
--- a/src/java.desktop/share/native/common/java2d/vulkan/VKBase.c
+++ b/src/java.desktop/share/native/common/java2d/vulkan/VKBase.c
@@ -43,9 +43,6 @@ static const uint32_t REQUIRED_VULKAN_VERSION = VK_MAKE_API_VERSION(0, 1, 2, 0);
 #define MAX_ENABLED_EXTENSIONS 5
 #define VALIDATION_LAYER_NAME "VK_LAYER_KHRONOS_validation"
 #define COUNT_OF(x) (sizeof(x)/sizeof(x[0]))
-#if defined(VK_USE_PLATFORM_WAYLAND_KHR)
-static struct wl_display *waylandDisplay;
-#endif
 
 static jboolean verbose;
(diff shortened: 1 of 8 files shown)


src/java.desktop/share/native/common/java2d/vulkan/VKBase.c
```diff
@@ -543,6 +550,16 @@ jboolean VK_FindDevices() {
         geInstance->vkGetPhysicalDeviceQueueFamilyProperties(
                 geInstance->physicalDevices[i], &queueFamilyCount, queueFamilies);
         int64_t queueFamily = -1;
+#if defined(VK_USE_PLATFORM_WAYLAND_KHR)
+        while (!wl_display_initialized) {
```
avu, line 554: Did you measure the performance gain from parallel initialization?

src/java.desktop/unix/native/libawt_wlawt/WLToolkit.c
```diff
@@ -741,6 +744,8 @@ Java_sun_awt_wl_WLToolkit_initIDs
   (JNIEnv *env, jclass clazz)
 {
     wl_display = wl_display_connect(NULL);
+    wl_display_initialized = 1;
+    syscall(SYS_futex, &wl_display_initialized, FUTEX_WAKE_PRIVATE, INT_MAX, NULL, NULL, 0);
```
avu, line 748: Why did you use such a kind of synchronization here?   (there is only one place in hotspot where JDK use it)

src/java.desktop/share/native/common/java2d/vulkan/VKBase.c
```diff
@@ -44,14 +44,16 @@ static const uint32_t REQUIRED_VULKAN_VERSION = VK_MAKE_API_VERSION(0, 1, 2, 0);
 #define VALIDATION_LAYER_NAME "VK_LAYER_KHRONOS_validation"
 #define COUNT_OF(x) (sizeof(x)/sizeof(x[0]))
 #if defined(VK_USE_PLATFORM_WAYLAND_KHR)
+#include <sys/syscall.h>
+#include <linux/futex.h>
+extern volatile uint32_t wl_display_initialized;
```
mkartashev, line 49: When inter-thread synchronization is spread between the native and Java worlds it becomes especially hard to follow. Maybe we can stay within Java in this instance? By the time `WLToolkit::initIDs()` exits normally (i.e. not through an exception), the connection to the Wayland server is guaranteed to exist. Possibly this synchronization point can be placed there?

mkartashev (changes requested):

src/java.desktop/share/native/common/java2d/vulkan/VKBase.c
avu: Yes, that's why I asked about it. Also, we want to extend the Vulkan pipeline to windows, so it's better to have a more general synchronization mechanism.

YaaZ: By the time `WLToolkit::initIDs()` exits normally, Vulkan should already be initialized, because it goes through `initIDs`->`notifyOutputConfigured` and creates a `WLGraphicsDevice`, which needs to know about the Vulkan support to create either `WLVKGraphicsConfig`, or `WLSMGraphicsConfig`. And Vulkan initialization needs a `wl_display` connection to query for presentation support, that's why the synchronization is placed here. This is only specific to Linux, as Windows does not depend on any external connection and therefore does not need such synchronization.

src/java.desktop/unix/native/libawt_wlawt/WLToolkit.c
```diff
@@ -741,6 +744,8 @@ Java_sun_awt_wl_WLToolkit_initIDs
   (JNIEnv *env, jclass clazz)
 {
     wl_display = wl_display_connect(NULL);
+    wl_display_initialized = 1;
```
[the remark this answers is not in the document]
YaaZ: Yes, but Vulkan initialization may already be waiting for the `wl_display`, so we need to notify it even if Wayland initialization failed. There is a null check in Vulkan code too.

src/java.desktop/unix/native/libawt_wlawt/WLToolkit.c
YaaZ: IDK, I just though futex is an easy, lightweight and already available tool for the case (waiting for the variable to change).

src/java.desktop/share/native/common/java2d/vulkan/VKBase.c
YaaZ: I didn't because I made these changes to solve the cyclic dependency in the first place. Original solution just crashes, so I cannot measure real gain, but I can get a rough estimate.

src/java.desktop/share/native/common/java2d/vulkan/VKBase.c
```diff
@@ -44,14 +44,13 @@ static const uint32_t REQUIRED_VULKAN_VERSION = VK_MAKE_API_VERSION(0, 1, 2, 0);
 #define VALIDATION_LAYER_NAME "VK_LAYER_KHRONOS_validation"
 #define COUNT_OF(x) (sizeof(x)/sizeof(x[0]))
 #if defined(VK_USE_PLATFORM_WAYLAND_KHR)
-extern struct wl_display *wl_display;
+static struct wl_display *wl_display;
```
mkartashev, line 47: This name clashes with `extern struct wl_display *wl_display` declared in WLToolkit.c
Maybe rename this one with a different prefix?

mkartashev (approved):

avu (approved):
Overal, LGTM. I think we'll get rid of global data later.

avu (approved):
LGTM

PR #399, opened 2024-06-04
JBR-7256 Vulkan: Implement FILL_PARALLELOGRAM primitive for flat color rendering

avu:
Implemented FILL_PARALLELOGRAM RQ primitive
Changed color shader to use push constant for color


YaaZ (approved):
Overall LGTM, but I'm thinking, maybe we could have a unified pipeline at least for colored primitive case, so that we could batch multiple primitives with different colors. Also as far as I can see, vertex buffers are being created and never destroyed, but this will be solved by buffer pools and proper resource management.

PR #404, opened 2024-06-11
JBR-5063 macOS: SIGILL at [libsystem_kernel] __kill in java.lang.IllegalStateException: Error - unable to initialize Metal after recreation of graphics device. Cannot load metal library...

avu:
Add fallback to MTLCreateSystemDefaultDevice for main display


YaaZ (approved):
LGTM, but in the future I think we need to respond to `CGDisplayRegisterReconfigurationCallback` events directly for adding/removing displays, like it's done in Wayland toolkit. I'm sure this way it will be way more robust.

PR #406, opened 2024-06-13
JBR-7305 Vulkan: Implement FILL_SPANS primitive for flat color rendering

avu:
Implemented flat color shape rendering


YaaZ (approved):
The only major difference between all plain-color rendering commands is the calculation of primitive coordinates, in the future we could refactor the common code, so that actual rendering function would look like:
1. `Vertex* verts = writeVertices(numVertices);`
2. Write coordinates into `verts`
3. `flushVertices();`

PR #410, opened 2024-07-01
JBR-6830 Poor performance with KDE Plasma 6 X11

mkartashev:
[JBR-6830](https://youtrack.jetbrains.com/issue/JBR-6830) Poor performance with KDE Plasma 6 X11

The X11 variant of KWin 6.* under certain circumstances starts sending dozens PropertyChange events for `NET_DESKTOP_GEOMETRY` per second. This slows things down not just for IDEA, but other X11 applications as well (probably because of requests issued by JBR in response to that event). So the real problem is with KWin.

This is a workaround until the KWin issue is fixed. It was verified by several users who reported back that the workaround worked for them.


YaaZ (approved):

mkartashev:
Merged into `main`

PR #411, opened 2024-07-07
JBR-7351 DCEVM: Respect explicit -XX:+ClassUnloading option

skybber:
Ensure that ClassUnloading is not disabled if the -XX:+ClassUnloading flag is explicitly set by the user.


mkartashev:
I'm not convinced that changing an undesirable, yet deterministic, behavior (code cache overflow) with even more undesirable (crash) and completely non-deterministic (may happen at random) is worth doing. Specifying `-XX:+ClassUnloading` does in no way imply that JVM might become unstable. Maybe create a totally new flag just for this purpose, but again, if it still makes the JVM unstable, why use it at all?

mkartashev (approved):

skybber:
Merged manually.

PR #415, opened 2024-07-11
JBR-5956 Provide more details on assertion failure

mkartashev:
[JBR-5956](https://youtrack.jetbrains.com/issue/JBR-5956) Provide more details on assertion failure

Use `JNU_RUNTIME_ASSERT(env, cond, msg)` defined in jni_util.h to crash JVM when `cond` is not true with the given message and source location information in the fatal error log.


vprovodin:
do we need a hotspot test checking this functionality?

mkartashev:
> do we need a hotspot test checking this functionality?

Not an easy thing to accomplish. I will look into this.

mkartashev:
I managed to add a test that verifies at least the core functionality. Please, have a look.

skybber (approved):

PR #421, opened 2024-07-23
JBR-7466 Exception on VM startup with `-Djava.util.zip.use.nio.for.zip.file.access=true`

knisht:
[JBR-7466](https://youtrack.jetbrains.com/issue/JBR-7466) Exception on VM startup with `-Djava.util.zip.use.nio.for.zip.file.access=true`

Follow-up for #413 : I encountered an exception during VM startup https://youtrack.jetbrains.com/issue/JBR-7392/Allow-using-NIO-in-ZipFile#focus=Comments-27-10185488.0-0 .

### DIFF
--- a/src/java.base/share/classes/java/util/zip/ZipFile.java
+++ b/src/java.base/share/classes/java/util/zip/ZipFile.java
@@ -114,7 +114,7 @@ public class ZipFile implements ZipConstants, Closeable {
     private final @Stable CleanableResource res;
 
     private static final boolean USE_NIO_FOR_ZIP_FILE_ACCESS =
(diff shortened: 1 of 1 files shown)


knisht:
Also fixes a problem mentioned in https://github.com/JetBrains/JetBrainsRuntime/pull/418#issue-2418273299

mkartashev:
The change itself looks good, but we need a new issue id in the commit message as this has already been ported to `jbr21` where force-push isn't allowed. Please, create a separate issue for this in youtrack.

mkartashev (approved):

PR #430, opened 2024-08-05
JBR-7459: [Wayland] Fixing too sensitive touchpad scrolling

NikitkoCent:
The patch fixes the problem described in [JBR-7459](https://youtrack.jetbrains.com/issue/JBR-7459):

>  From [JBR-5673](https://youtrack.jetbrains.com/issue/JBR-5673/Wayland-support-touch-scrolling):
> 
> > [...] the number of pixels to be scrolled on IDEA under pure Wayland [...] depends [...] on the movement speed: the more slowly the fingers move, the more pixels are scrolled. The difference can reach 14 times!
> 
> Usually it leads to the feeling that the touchpad is too sensitive at scrolling.

I recommend to review the patch commit-by-commit (I think this will help to understand how the changes are related to each other). Each commit solves one particular sub-task, which is described in the commit message. After the review all the commits will be squashed.

### DIFF
--- a/src/java.desktop/unix/classes/sun/awt/wl/WLComponentPeer.java
+++ b/src/java.desktop/unix/classes/sun/awt/wl/WLComponentPeer.java
@@ -102,6 +102,9 @@ public class WLComponentPeer implements ComponentPeer {
             {"move"}, // MOVE_CURSOR
     };
 
+    // Changing this constant may require adjustments in convertAxisVectorToPreciseWheelRotations
+    private static final int WHEEL_SCROLL_AMOUNT = 3;
+
     private static final int MINIMUM_WIDTH = 1;
     private static final int MINIMUM_HEIGHT = 1;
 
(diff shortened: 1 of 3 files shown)


src/java.desktop/unix/native/libawt_wlawt/WLToolkit.c
```diff
@@ -313,7 +340,9 @@ static const struct wl_pointer_listener wl_pointer_listener = {
         .frame         = wl_pointer_frame,
         .axis_source   = wl_pointer_axis_source,
         .axis_stop     = wl_pointer_axis_stop,
-        .axis_discrete = wl_pointer_axis_discrete
+        .axis_discrete = wl_pointer_axis_discrete/*,
+        This is only supported if the libwayland-client supports version 8 of the wl_pointer interface
+        .axis_value120 = wl_pointer_axis_value120*/
```
mkartashev, line 345: We can achieve this by compiling against version 8, but binding to whatever is available; see, for example, xdg_wm_base in `registry_global()`: https://github.com/JetBrains/JetBrainsRuntime/blob/045a7e4e51232a6b0522fa274608d4a23405332f/src/java.desktop/unix/native/libawt_wlawt/WLToolkit.c#L504

src/java.desktop/unix/classes/sun/awt/wl/WLComponentPeer.java
```diff
@@ -102,6 +102,9 @@ public class WLComponentPeer implements ComponentPeer {
             {"move"}, // MOVE_CURSOR
     };
 
+    // Changing this constant may require adjustments in convertAxisVectorToPreciseWheelRotations
+    private static final int WHEEL_SCROLL_AMOUNT = 3;
```
mkartashev, line 106: There are now two parameters that control the scrolling speed. I'd rather we have one or at least have both near one another. If having this one is absolutely necessary, please make it into a method similar to `convertAxisVectorToPreciseWheelRotations`.

src/java.desktop/unix/classes/sun/awt/wl/WLComponentPeer.java
```diff
@@ -1152,19 +1154,52 @@ void dispatchPointerEventInContext(WLPointerEvent e, WLInputState oldInputState,
...
-                    MouseEvent.MOUSE_WHEEL,
-                    timestamp,
-                    newInputState.getModifiers(),
-                    x, y,
-                    xAbsolute, yAbsolute,
-                    1,
-                    isPopupTrigger,
-                    MouseWheelEvent.WHEEL_UNIT_SCROLL,
-                    1,
-                    Integer.signum(e.getAxis0Value()) * WHEEL_SCROLL_AMOUNT);
```
NikitkoCent, line 1166: Basically the root of the issue was in this line, I believe. When you move your fingers on the touchpad slowly, Wayland generates lots of `wl_pointer::axis` events but with really small values, like 0.2, and we treated them as a full wheel step.

mkartashev (approved):

src/java.desktop/unix/classes/sun/awt/wl/WLComponentPeer.java
NikitkoCent: > I'd rather we have one or at least have both near one another. If having this one is absolutely necessary, please make it into a method similar to `convertAxisVectorToPreciseWheelRotations`

Yeah, I placed all the constants together in #429

src/java.desktop/unix/native/libawt_wlawt/WLToolkit.c
NikitkoCent: > compiling against version 8

Most likely it'll require updating our building agents.

BTW, are we going to have a policy for upgrading libwayland-client and/or other dependencies of extension protocols? Like whether we're going to keep backward-compatibility at compile-time with older version of dependencies and so on. I've filed JBR-7463 to investigate such questions.

mkartashev: > Most likely it'll require updating our building agents.

Yes, probably worth doing anyway as there were bugs reported aginst older versions of wayland-scanner, etc.

> BTW, are we going to have a policy for upgrading libwayland-client and/or dependencies of extension protocols?

I think it's too early to have a policy us being are in an active development phase right now.

tsarn (approved):

t0rr3sp3dr0:
@NikitkoCent, @mkartashev, and @tsarn, can we get this merged or is there something blocking it?

NikitkoCent:
Merged (https://youtrack.jetbrains.com/issue/JBR-7459/Wayland-touchpad-scrolling-is-too-sensitive#focus=Comments-27-10656634.0-0).

PR #434, opened 2024-08-12
JBR-7481: Work around for mouse entered/exited events

stachenov:
Implement a workaround for the bug when mouse entered/exited events are not generated by AppKit for whatever reason. Needed for the cursor set using `setCursor` to be respected, as it relies on `sun.lwawt.LWWindowPeer#getPeerUnderCursor`, which is updated according to mouse events. If it's wrong, the cursor is also wrong. See the YT issue for details:

https://youtrack.jetbrains.com/issue/JBR-7481


src/java.desktop/macosx/classes/sun/lwawt/MouseEnteredExitedBugWorkaround.java
```diff
@@ -0,0 +1,133 @@
...
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+package sun.lwawt;
+
+import java.awt.event.MouseEvent;
+
+class MouseEnteredExitedBugWorkaround {
```
NikitkoCent, line 30: Was there a reason to split the workaround's logic between `Jbr7481MouseEnteredExitedFix` and `MouseEnteredExitedBugWorkaround`? I feel like keeping everything in `Jbr7481MouseEnteredExitedFix` is better from the responsibility point of view - currently it's not obvious which problem `MouseEnteredExitedBugWorkaround` solves and what it belongs to (ofc until you start exploring and git-blaming the code)

OnePatchGuy (changes requested):
Thank you for the patch! I've tested it and found no problems, however there's 1 question

src/java.desktop/macosx/classes/sun/lwawt/MouseEnteredExitedBugWorkaround.java
stachenov: Makes sense. Updated and force-pushed.

NikitkoCent:
Looks good. I'll merge it into `main` and wait for the post-commit testing results. If they're fine, the patch will also be delivered to `jbr17` and `jbr21`. Until then let's keep the PR open (so that I won't forget to finish all the work).

OnePatchGuy (approved):

PR #435, opened 2024-08-12
JBR-7313 Wayland: error: xdg_surface buffer does not match the configured maximized state

mkartashev:
[JBR-7313](https://youtrack.jetbrains.com/issue/JBR-7313) Wayland: error: xdg_surface buffer does not match the configured maximized state

The most important part is to keep the surface size that came from the Wayland server in the 'configure' event and pass it to `nativeSetSurfaceSize()` and `nativeSetWindowGeometry()` unmodified, especially when in maximized state.

It seems that passing the scale to `SurfaceData` is unnecessary as it is not used in `WLSMSurfaceData`. If it is safe to remove the scale from `WLVKSurfaceData`, let's do that and free SurfaceData from worrying about scaling altogether.

### DIFF
--- a/src/java.desktop/unix/classes/sun/awt/wl/WLComponentPeer.java
+++ b/src/java.desktop/unix/classes/sun/awt/wl/WLComponentPeer.java
@@ -73,7 +73,6 @@ import java.awt.image.ColorModel;
 import java.awt.image.VolatileImage;
 import java.awt.peer.ComponentPeer;
 import java.awt.peer.ContainerPeer;
-import java.awt.peer.KeyboardFocusManagerPeer;
 import java.util.ArrayList;
 import java.util.Objects;
 import java.util.function.Supplier;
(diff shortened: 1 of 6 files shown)


src/java.desktop/unix/classes/sun/awt/wl/WLComponentPeer.java
```diff
@@ -1443,4 +1441,94 @@ static Window getNativelyFocusableOwnerOrSelf(Component component) {
         }
         return result;
     }
+
+    private class WLSize {
+        private final Dimension javaSize = new Dimension(); // in the client (Java) space, protected by dataLock
+        private final Dimension pixelSize = new Dimension(); // in pixels, protected by dataLock
+        private final Dimension surfaceSize = new Dimension(); // in surface units, protected by dataLock
```
YaaZ, line 1448: Could you add more details/examples here? It still seems not clear to me, what is the difference between pixel and surface size

YaaZ (approved):

src/java.desktop/unix/classes/sun/awt/wl/WLComponentPeer.java
mkartashev: I added Javadoc for all these fields, but their relationships with one another and various scales is best learned from the code itself, the `derive*` methods of that class.

PR #443, opened 2024-08-21
JBR-7517 build JBR artefacts with CDS archives

vprovodin:


vprovodin:
The failure of the test jb/build/CDSArchivesTest.sh on Windows-aarch64 is expected. It successfully passess in other configurations

PR #445, opened 2024-08-21
JBR-7016 IDEA 2024.2 Wayland: UI Crash when selecting Code and pressing Alt+Enter

mkartashev:
[JBR-7016](https://youtrack.jetbrains.com/issue/JBR-7016) IDEA 2024.2 Wayland: UI Crash when selecting Code and pressing Alt+Enter

Account for the possibility of popups having another popup as their parent.

### DIFF
--- a/src/java.desktop/unix/classes/sun/awt/wl/WLComponentPeer.java
+++ b/src/java.desktop/unix/classes/sun/awt/wl/WLComponentPeer.java
@@ -271,12 +271,40 @@ public class WLComponentPeer implements ComponentPeer {
     private static Window getToplevelFor(Component component) {
         Container container = component instanceof Container c ? c : component.getParent();
         for(Container p = container; p != null; p = p.getParent()) {
-            if (p instanceof Window) {
-                return (Window)p;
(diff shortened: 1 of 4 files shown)


src/java.desktop/unix/classes/sun/awt/wl/WLComponentPeer.java
```diff
@@ -271,14 +271,42 @@ public boolean requestFocus(Component lightweightChild, boolean temporary,
...
+            return c.getLocation();
+        }
+
+        int x = 0, y = 0;
+        while (c != null) {
+            if (c instanceof Window window) {
+                // The location of non-popup windows has no relevance since
+                // there are no absolute coordinates in Wayland.
+                // The popup windows position, on the other hand, is set relative to their
+                // parent toplevel.
```
YaaZ, line 295: So if we have structure like this: `TopLevelWindow` -> `Popup1` -> `Popup2`
Will `Popup2` be positioned relative to `TopLevelWindow` and not `Popup1`?
Sounds kinda counter-intuitive.

YaaZ (approved):

PR #453, opened 2024-08-30
JBR-7579 Fix SurfaceManager.cacheMap retaining strong references.

YaaZ:


YaaZ:
I'm not sure that having value as a `WeakReference` is the right solution.
However in our case, leaked `MTLGraphicsContext` is reachable from both key and value, like this:
`key` -> `MTLGraphicsConfig`
`value` -> `MTLSurfaceDataProxy` -> `MTLGraphicsConfig`
But I don't think keeping `MTLGraphicsConfig` as a weak ref inside `MTLSurfaceDataProxy` is more correct either.

When I come with this patch to OpenJDK, they will ask me: what is the point of this cache, when its values must be duplicated elsewhere in order to not be collected?

So the ideal solution I would want is something like this:
- if key (GraphicsConfig) is strongly reachable ONLY via the value (SurfaceDataProxy), then collect the whole node
- otherwise keep the value (SurfaceDataProxy) strongly reachable

But I cannot currently figure out a way to do this.

bourgesl:
Looks like a great puzzle on reference... could you elaborate who is responsible to keep alive this references ?
I can maybe help here...

YaaZ:
So originally GraphicsConfigs were getting stuck in this cache, preventing them from being destroyed.

GraphicsConfigs are usually used as keys in this map, so the first thing we need to do is to make the map a WeakHashMap. But this alone is not enough, because as values we store SurfaceDataProxies, which in turn have a reference to the same GraphicsConfig, thus retaining a strong reference to it. There's even a note about this in the documentation of WeakHashMap:

> Implementation note: The value objects in a WeakHashMap are held by ordinary strong references. Thus care should be taken to ensure that value objects do not strongly refer to their own keys, either directly or indirectly, since that will prevent the keys from being discarded. Note that a value object may refer indirectly to its key via the WeakHashMap itself; that is, a value object may strongly refer to some other key object whose associated value object, in turn, strongly refers to the key of the first value object. If the values in the map do not rely on the map holding strong references to them, one way to deal with this is to wrap values themselves within WeakReferences before inserting, as in: m.put(key, new WeakReference(value)), and then unwrapping upon each get.

So I wrapped values into WeakRefs. However now this means that when device is stll alive, but there are no more strong refs to the SurfaceDataProxy, it will be GCed, which may be undesirable (?).

So the main question is: is there a code, which relies on this cache keeping SurfaceDataProxies alive, or are they already strongly referenced as long as they needed and there is nothing to worry?
Does this generally apply to any kind of data that can be stored in that cache? The only use case I could find is that GraphicsConfig + SurfaceDataProxy one, but I may have missed something.

I will investigate it a bit more and post an update.

YaaZ:
Even if we try to keep GraphicsConfig as a weak ref inside all implementations of SurfaceDataProxy, it will still be retained via SurfaceDataProxy.cachedSD. We cannot make it weak as well, because it basically "owns" the SurfaceData and should not be collected as long as SurfaceDataProxy is reachable. Descending deeper into SurfaceData and getting rid of strong refs to GraphicsConfiguration there is just too much of refactoring, and too many places where something can go wrong.

I would say, we have stumbled upon limitations of WeakHashMap implementation and it doesn't seem currently possible to implement it "the right way", that is, retain value as long as key is reachable, but do not let it prevent the key from being collected. This would require some GC magic to collect the whole node at once, like making value a weak ref, but adding a virtual "key->value" edge into reachability graph.

But as a workaround, key can itself serve as a node - "physically" contain corresponding values. This way strong refs from the value would not prevent the key from being collected. I will investigate this approach.

bourgesl:
Reference can have their own ReferenceCleaner... so you could handle objects when the reference becomes weakly reachable.
I used special Cleaner in marlin to free reliabilly off-heap resources, see:
https://github.com/bourgesl/marlin-renderer/blob/jdk/src/main/java/sun/java2d/marlin/OffHeapArray.java

YaaZ:
But Cleaner wouldn't help, because GraphicsConfiguration would never become weakly reachable due to the way this cache is organized.

bourgesl:
Maybe you could introduce a special object, like GraphicsConfigRef ou SurfaceRef that could be handled by special cleaner... the big question here is: who (which object) is supposed to make surface & graphicsConfig alive...

YaaZ:
Now I'm pretty sure this is the best solution.
1. I moved proxy cache into a separate object. Now caching level is determined by placement of this cache object instead of a "proxy key". This cache "owns" proxies and therefore strong refs via them do not prevent the cache (and hence GraphicsConfig) from being detected as weakly reachable.
2. I removed a global JNI ref to the GraphicsConfig from native MTLSurfaceData, it was not needed there. This code was blindly copied from the OGL implementation and therefore this bug can affect OGL too, I will investigate it later.

src/java.desktop/macosx/classes/sun/java2d/metal/MTLSurfaceData.java
```diff
@@ -623,13 +623,13 @@ public Object getDestination() {
...
      */
-     public static void dispose(long pData, MTLGraphicsConfig gc) {
+     public static void dispose(long pData) {
         MTLRenderQueue rq = MTLRenderQueue.getInstance();
         rq.lock();
         try {
             // make sure we have a current context before
             // disposing the native resources (e.g. texture object)
-            MTLContext.setScratchSurface(gc);
+            MTLContext.setScratchSurface(0);
```
avu, line 632: I think it's incorrect. We need to have a valid context to dispose a surface. So, we just skip MTLSD_Delete because of this code.

src/java.desktop/macosx/classes/sun/java2d/opengl/CGLGraphicsConfig.java
```diff
@@ -110,8 +112,8 @@ private CGLGraphicsConfig(CGraphicsDevice device, long configInfo,
     }
 
     @Override
-    public Object getProxyKey() {
```
avu, line 113: Why don't you leave the signature as is and return your specific object? Maybe I'm missing something but the change is big just because of such unnecessary refactoring.

avu (changes requested):

src/java.desktop/macosx/classes/sun/java2d/metal/MTLSurfaceData.java
YaaZ: I don't see where do we need a valid context to dispose a surface. `mtlc` is not actually used for destroying the surface, only to free glyph caches and commit the command buffer. I achieved the same behavior by setting scratch surface to 0.
We won't skip MTLSD_Delete because I removed `CONTINUE_IF_NULL(mtlc);` statement from `DISPOSE_SURFACE`. That actually looked incorrect to me - we would leak surface resources just because we don't have a current context, which is again, not actually used for surface destruction.

src/java.desktop/macosx/classes/sun/java2d/opengl/CGLGraphicsConfig.java
YaaZ: What do you mean by "your specific object"? This whole refactoring was motivated by the need to physically contain the cache inside of the GraphicsConfiguration (in most implementations).

src/java.desktop/macosx/classes/sun/java2d/metal/MTLSurfaceData.java
YaaZ: Actually, now that I look at the code again, we don't actually seem to need to do `SET_SCRATCH_SURFACE` at all - glyph caches will be freed and command buffer will be committed anyway by `DISPOSE_SURFACE`. Moreover, this is even preferred, as `DISPOSE_SURFACE` will wait for completion and prevent early destrution of surface resources which may still be in use.

src/java.desktop/macosx/classes/sun/java2d/opengl/CGLGraphicsConfig.java
avu: I mean that you don't need to change the signature of this method to introduce your new approach (cache inside the GC).  So, you don't need to change 22 file with just signature change

avu: OK, I see the actual usage of the caches, so Object type will not work. But why you put the caches into child classes of  CGraphicsConfig (MTLGraphicsConfig and CGLGraphicsConfig) but not into the base class itself?

src/java.desktop/macosx/native/libawt_lwawt/java2d/metal/MTLRenderQueue.m
```diff
@@ -639,7 +640,11 @@ void MTLRenderQueue_CheckPreviousOp(jint op) {
                     jlong pData = NEXT_LONG(b);
                     BMTLSDOps *mtlsdo = (BMTLSDOps *)jlong_to_ptr(pData);
                     if (mtlsdo != NULL) {
-                        CONTINUE_IF_NULL(mtlc);
+                        if (mtlc != NULL) {
```
avu, line 643: After your setScratchSurface(0) this code will not be executed because of mtlc=NULL

avu (changes requested):

src/java.desktop/macosx/classes/sun/java2d/metal/MTLSurfaceData.java
avu: OK, looks like we don't need gc anymore, but setting the scratch surface to 0 effectively disables font cache cleanup code that you've added into DISPOSE_SURFACE case.

src/java.desktop/macosx/classes/sun/java2d/opengl/CGLGraphicsConfig.java
bourgesl: It looks like the role or objectives of setSurface, setScratchSurface, disposeSurface, flushSurface, flushBuffer, disposeConfig... are not well-known and quite specific to each pipeline. 

What transitions are allowed between these states (it is a logical state machine) ?
I like the new flush buffer corresponding to the frame boundary... but flushBuffer sounds different => flushFrame or flushDisplay....

src/java.desktop/macosx/classes/sun/java2d/metal/MTLSurfaceData.java
YaaZ: I'm going to just remove `SET_SCRATCH_SURFACE` call. Cache will be cleaned by `DISPOSE_SURFACE` with proper wait.

src/java.desktop/macosx/classes/sun/java2d/opengl/CGLGraphicsConfig.java
YaaZ: Indeed, I could put the cache into `CGraphicsConfig`, will fix this.

bourgesl (approved):

src/java.desktop/unix/classes/sun/awt/X11GraphicsDevice.java
```diff
@@ -36,12 +36,10 @@
 import java.awt.Window;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Objects;
+import java.util.*;
```
avu, line 39: Do not replace explicit imports with wild cards

avu (changes requested):

src/java.desktop/unix/classes/sun/awt/X11GraphicsDevice.java
YaaZ: That was automatic IDEA refactoring, fixed.

PR #455, opened 2024-09-03
JBR-7484 Update the cursor on mouse entered/exited

stachenov:
AppKit resets the cursor on native mouse entered/exited events. Depending on the order of events, it may end up setting the wrong cursor. So update it forcibly on such events.


OnePatchGuy (approved):
Cool, thanks for the patch! Like for #434, let's see how it goes on `main` and then I'll deliver it to `jbr17`, `jbr21` as well.

UPD: pushed to `main`.

NikitkoCent:
Everything seems fine, the fix has been delivered to jbr17, jbr21 as well.

PR #458, opened 2024-09-05
JBR-4725 File dialog modality

YaaZ:


avu (approved):
LGTM, however it would be nice to have a reg test

YaaZ:
Actually, there is a related test already: `FileDialogNonModal7Test`. It fails before and passes after my change, I will consider using this as an argument to port the change into OpenJDK, possibly extending the test set.

PR #484, opened 2024-12-10
Update JavadocFormatterTest.java

fhar28:


mkartashev:
Hello! Could you provide some explanation for this patch, please? What problem does it solve and how? Why do you wish to merge it into branch `17_250` rather than `main`?
Thanks!

PR #486, opened 2024-12-19
JBR-8043 Fix CSE from Files.probeContentType with overridden DefaultFileSystemProvider

vladimirlagunov:
Before this commit, the code `DefaultFileTypeDetector` implied that the default file system provider is always an instance of `UnixFileSystemProvider`. It's not true when `-Djava.nio.file.spi.DefaultFileSystemProvider` is specified.

In general, there's a possibility to specify a custom `FileTypeDetector` and define it via service loading. (See `java.nio.file.Files.probeContentType`). However, the problematic code is invoked during the static initialization of the class `java.nio.file.Files.FileTypeDetectors`, which is also the entry point for the service loader. This way, it's even impossible to define a custom `FileTypeDetector`.

This commit replaces the strict class cast with a conditional one, and the new code returns a no-op file type detector in cases when the default file system provider is overridden.

### DIFF
--- a/src/java.base/unix/classes/sun/nio/fs/DefaultFileTypeDetector.java
+++ b/src/java.base/unix/classes/sun/nio/fs/DefaultFileTypeDetector.java
@@ -25,7 +25,9 @@
 
 package sun.nio.fs;
 
+import java.io.IOException;
 import java.nio.file.FileSystems;
+import java.nio.file.Path;
 import java.nio.file.spi.FileTypeDetector;
 import java.nio.file.spi.FileSystemProvider;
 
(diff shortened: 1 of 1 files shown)


src/java.base/unix/classes/sun/nio/fs/DefaultFileTypeDetector.java
```diff
@@ -34,6 +36,14 @@ private DefaultFileTypeDetector() { }
 
     public static FileTypeDetector create() {
         FileSystemProvider provider = FileSystems.getDefault().provider();
-        return ((UnixFileSystemProvider)provider).getFileTypeDetector();
+        if (provider instanceof UnixFileSystemProvider) {
```
mkartashev, line 39: You can avoid casting on the next line if you bind the instance to a variable like so
```
if (provider instanceof UnixFileSystemProvider unixProvider) {
   return unixProvider.getFileTypeDetector();
}
```

mkartashev (changes requested):

src/java.base/unix/classes/sun/nio/fs/DefaultFileTypeDetector.java
vladimirlagunov: https://github.com/JetBrains/JetBrainsRuntime/pull/486/commits/18cb31b8b75dce871926cba72d9e4c66809315a1

mkartashev (approved):
Looks good. I will add a dedicated test, run the existing tests, and merge this manually (first into `main` and then into `jbr21`).

mkartashev:
Merged into `main` and `jbr21` ([fd29e06c9c](https://github.com/JetBrains/JetBrainsRuntime/commit/fd29e06c9c098322fd0e7ae2a4e8d5ae46564c0e)).

PR #492, opened 2025-01-13
vpr/jbr21_JBR-8111_smooth_transition

vprovodin:


vprovodin:
Not sure that we need to have these changes in jbr21.
These two commits are intended to be cherry-picked always when a new created branch targeted for 2025.1 is created till FF. After FF the commits will be landed in the 2025.1 branch and the cherry-picks won't be required.

mkartashev (approved):

PR #497, opened 2025-01-20
JBR-8156 restoring WSL visibility in the folder picker mode

trespasserw:


OnePatchGuy (commented):
One thing worries me, though I'm not a specialist in file dialogs: the patch changes behavior for all directory-only file dialogs. Are there cases when we have to only show "native" directories, so that this change will bring us a regression?
Alternative approach would be to extend JBR API to be able to exclude `FOS_FORCEFILESYSTEM` only when it's really required.

UPD: besides that, the patch looks good to me

trespasserw:
> Are there cases when we have to only show "native" directories, so that this change will bring us a regression? Alternative approach would be to extend JBR API to be able to exclude `FOS_FORCEFILESYSTEM` only when it's really required.

Found nothing so far. The behavior of the flag is really buggy: directories inside WSL are file-like and can be chosen, the same applies to UNC paths - yet the "Network" node is visible and the "Linux" one is not. Moreover, one can trick the dialog unto showing that node by pinning any WSL directory to the quick access list and navigating to its root -- then the "Linux" node magically appears! Finally, it only behaves this way when `FOS_PICKFOLDERS` is set. Since this is a bug, I don't think changes in the JBR API are needed.

avu (approved):
Looks good for me

trukhinyuri:
@OnePatchGuy 
Here’s how it works: The FOS_PICKFOLDERS flag in _FILEOPENDIALOGOPTIONS allows the Open or Save dialogs (Win32 API) to select only folders (or objects provided by SHBrowseForFolderA, such as the “This PC” node).

Meanwhile, \\wsl$ in the WinAPI is considered a “filesystem ancestor” (not a regular folder but a virtual one) and uses its own navigation mechanisms. This is determined by the SFGAO_FILESYSANCESTOR flag. Objects that set this flag typically appear when FOS_PICKFOLDERS is enabled. However, WSL, DevDrive, and several other similar nodes do not show up in classic dialogs.

This behavior is correct, because these entities aren’t treated by the system as fully-fledged “folders” and don’t provide the standard SHBrowseForFolderA logic. [Raymond Chen](https://devblogs.microsoft.com/oldnewthing/20231011-00/?p=108884)—one of the WinAPI architects—writes about this in detail.

Why is it done this way? Microsoft deliberately restricts the classic Win32 API from accessing “semi-virtual” folders that provide broader or non-standard navigation methods (WSL, DevDrive, or nodes attached via drivers for unofficial Windows file systems). This is needed for compatibility: legacy applications that are unaware of pseudo-virtual paths like \\wsl$ can break if they try to handle them according to regular Win32 rules, since permissions, symlinks, or case sensitivity may work differently there.

It’s not a bug but a feature—implemented for compatibility with classic applications. 
**For our needs, it doesn’t seem necessary; we’re not dealing with classic approaches just yet.**

trespasserw:
In theory, the removal of `FOS_FORCE­FILE­SYSTEM` can result in previously invisible nodes to appear in the dialog. That's why I need to add additional validation (I'm on it, but the code is complicated, so it will take some time).

But the behavior is definitely a bug. The Chen's article confirms this by saying, "they may need to pass through these unpickable folders in order to get something that they can pick." Which makes sense, and is further corroborated by the fact that without `FOS_PICKFOLDERS` the "Linux" node is visible.

PR #500, opened 2025-01-21
vpr/main_JBR-8161_remove_docker_files

vprovodin:


.github/README.md
```diff
@@ -116,28 +116,29 @@ coverage of all the details.
...
-$ cd jb/project/docker
-$ ./mkdocker_x86_64.sh
-...
-Successfully built 942ea9900054
+$ docker pull jetbrains/runtime:oraclelinux8_aarch64
 ```
-Run these commands in the new container:
+or
+```
+$ docker pull jetbrains/runtime:oraclelinux8_x64
```
mkartashev, line 125: OEL containers require a support contract, no? Maybe it's best to just drop the docker section for Linux?

.github/README.md
```diff
@@ -116,28 +116,29 @@ coverage of all the details.
...
 ```
 
 ### Ubuntu Linux
-Install the necessary tools, libraries, and headers with:
-```
-$ sudo apt-get install autoconf make build-essential libx11-dev libxext-dev libxrender-dev libxtst-dev \
-       libxt-dev libxrandr-dev libcups2-dev libfontconfig1-dev libasound2-dev libspeechd-dev libwayland-dev \
-       wayland-protocols libxkbcommon-x11-0
-```
+
```
mkartashev, line 141: Why removing these hints?

.github/README.md
vprovodin: No. The images from DockerHub (with the prefix jetbrains/runtime )  do not have support contract. OL8 images from DockerHub are covered by Basic support

.github/README.md
vprovodin: They were ooutdated - IMHO it would better to have a proper image for building rather instructions in READ.ME that need to be regularly updated...

mkartashev: Then let's at least keep the "Install the necessary tools, libraries, and headers" bit.

mkartashev (approved):

PR #509, opened 2025-02-05
JBR-8235 Don't throw ArrayIndexOutOfBoundsException from FileSystemProvider.newByteChannel on Unix

vladimirlagunov:
[JBR-8235](https://youtrack.jetbrains.com/issue/JBR-8235) AIOOBE from newByteChannel on empty path name

Due to a missed check, a call of `newByteChannel(Path.of(""), EnumSet.of(WRITE, CREATE_NEW))` led to `ArrayIndexOutOfBoundsException`, because there was no check for an empty path.

Now this code throws `FileAlreadyExists`, as it happens for other corner cases like `Path.of(".")"`


mkartashev:
Could you please explain the need for this fix and why does the new code assumes that Path.of("") refers to the current directory? 
Also, is Unix's file channel the only place where such fix is required?

vladimirlagunov:
> the need for this fix

`ArrayIndexOutOfBoundException` isn't specified in the JavaDoc as a possible exception. Unlike `IllegalArgumentException`, `Error` and so on, AIOOBE has never been treated as an implicit API of every method. So, AIOOBE should be treated as an unexpected internal error.

> why does the new code assumes that Path.of("") refers to the current directory?

Because `Path.of("").toAbsolutePath()` returns the current directory, as well as `Files.list(Path.of(""))` returns files in the current directory, and so on.

> is Unix's file channel the only place where such fix is required?

This example shows that WindowsFileSystemProvider doesn't have a similar issue.

```java
import java.io.IOException;
import java.nio.file.FileSystemException;
import java.nio.file.FileSystems;
import java.nio.file.Path;

import java.nio.file.StandardOpenOption;
import java.util.List;
import java.util.ArrayList;
import java.util.EnumSet;
import java.util.Set;

class Scratch {
    public static void main(String[] args) throws IOException {
        List<Set<StandardOpenOption>> optionsList = new ArrayList<>();
        {
            StandardOpenOption[] values = StandardOpenOption.values();
            for (int i = 0; i < (1 << values.length); i++) {
                Set<StandardOpenOption> combination = EnumSet.noneOf(StandardOpenOption.class);
                for (int j = 0; j < values.length; j++) {
                    if ((i & (1 << j)) != 0) {
                        combination.add(values[j]);
                    }
                }
                optionsList.add(combination);
            }
        }

        for (var options : optionsList) {
            System.out.println(options);
            try {
                FileSystems.getDefault().provider().newByteChannel(Path.of(""), options).close();
            } catch (IllegalArgumentException | FileSystemException ignored) {
                // Nothing.
            } catch (Throwable e) {
                e.printStackTrace();
            }
        }
    }
}
```

mkartashev:
OK, we can keep this fix for now, but I will file an OpenJDK issue as
1. I'm not sure Path.of("") should always represent the current directory (maybe some specification needs to be patched?)
2. That the file channel is the only place that is affected; I understand that the exact same issue is not present on Windows, but there is more to NIO than just FileChannel

mkartashev (approved):

vladimirlagunov:
> I'm not sure Path.of("") should always represent the current directory (maybe some specification needs to be patched?)

I found a written proof: https://github.com/JetBrains/JetBrainsRuntime/blob/main/src/java.base/share/classes/java/nio/file/Path.java#L48-L50

mkartashev:
> > I'm not sure Path.of("") should always represent the current directory (maybe some specification needs to be patched?)
> 
> I found a written proof: https://github.com/JetBrains/JetBrainsRuntime/blob/main/src/java.base/share/classes/java/nio/file/Path.java#L48-L50

Great, thank you!

mkartashev:
Merged into `main`.

PR #513, opened 2025-02-10
JBR-8118 TextureWrapperImage for MTLTexture

ujpv:


ujpv:
JBR-API
https://github.com/JetBrains/JetBrainsRuntimeApi/pull/13

src/java.desktop/macosx/classes/sun/java2d/metal/MTLSurfaceData.java
```diff
@@ -674,4 +681,46 @@ protected void loadNativeRaster(long pRaster, int width, int height, long pRects
...
+        private final Image myImage;
+
+        private MTLTextureWrapperSurfaceData(MTLGraphicsConfig gc, Image image, long pTexture) throws IllegalArgumentException {
+            super(null, gc, ColorModel.getRGBdefault(), RT_TEXTURE, /*width=*/ 0, /*height=*/ 0);
+            myImage = image;
+
+            MTLRenderQueue rq = MTLRenderQueue.getInstance();
+            rq.lock();
+
+            AtomicBoolean success = new AtomicBoolean(false);
```
mkartashev, line 698: Maybe move this before `rq.lock()` so that it doesn't leave the render queue locked in case of OOME, for example.

src/java.desktop/share/classes/com/jetbrains/desktop/image/TextureWrapperImage.java
```diff
@@ -0,0 +1,143 @@
...
+        } catch (Exception e) {
+            throw new InternalError("Unexpected exception during reflection", e);
+        }
+
+        return 0;
+    }
+
+    @Override
+    public int getWidth(ImageObserver observer) {
+        return (int) (sd.getBounds().width / sd.getDefaultScaleX());
```
mkartashev, line 99: I believe it's best to scale the size up and coordinates down. This way you automatically avoid zero size problems for small images, among other things.

avu (changes requested):

src/java.desktop/share/classes/com/jetbrains/desktop/image/TextureWrapperImage.java
```diff
@@ -0,0 +1,143 @@
...
+        try {
+            if (isInstanceOf(gc, "sun.java2d.metal.MTLGraphicsConfig")) {
+                return 1;
+            }
+        } catch (Exception e) {
+            throw new InternalError("Unexpected exception during reflection", e);
+        }
+
+        return 0;
+    }
```
YaaZ, line 95: What concerns me is that the service wil always be considered supported because it has an implementation in shared code. But still it will only work if our GC is a Metal one. Maybe better make a class, like `SharedTexturesService` and mark it with `@JBRApi.Provides("SharedTextures")`, make `wrapTexture` and `getTextureType` its instance methods (without annotations) and check supported texture type according to the current toolkit once in a constructor, if there is no supported texture type, throw `JBRApi.ServiceNotAvailableException` from the service constructor.

ujpv: fixed

src/java.desktop/unix/classes/sun/java2d/UnixSurfaceManagerFactory.java
```diff
@@ -73,4 +75,8 @@ public VolatileSurfaceManager createVolatileManager(SunVolatileImage vImg,
         }
     }
 
+    @Override
+    public SurfaceManager createTextureWrapperSurfaceManager(GraphicsConfiguration gc, Image image, long texture) {
+        throw new UnsupportedOperationException("unsupported");
```
mkartashev, line 80: It's not really important, but UOE already means that something is unsupported, so there's no need for a generic message like this. Same goes for WindowsSurfaceManagerFactory.java

mkartashev (changes requested):

src/java.desktop/share/classes/com/jetbrains/desktop/SharedTextures.java
```diff
@@ -0,0 +1,85 @@
...
+        } catch (Exception e) {
+            throw new InternalError("Unexpected exception during reflection", e);
+        }
+
+        return 0;
+    }
+
+    private static boolean isInstanceOf(Object obj, String className) {
+        try {
+            var clazz = Class.forName(className);
```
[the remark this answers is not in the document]
ujpv: Thanks!

mkartashev (approved):

PR #518, opened 2025-02-25
JBR-8332 SizeRequirements: Compensate for remaining unallocated span when laying out HTML tables

piotrtomiak:
For `jbr21` and `main` branches


vprovodin:
*Commit testing* was executed on *main* + PR, no issues were detected

mkartashev:
Are there any existing tests covering this change? If not, please create a JTREG test that doesn't pass without this change and passes with it.

piotrtomiak:
@mkartashev - I've tried to create a JTREG test now, but I've failed to reproduce the problem. It looks like I've managed to pinpoint the issue on the IJ side. This logic here looks fishy, but since there is no visible repro state in JDK, I will close this PR and apply a fix on IJ side.

PR #525, opened 2025-03-18
JBR-8435 Provide legal info to JBR builds

vprovodin:


mkartashev:
Will this new file be packaged into all the distributions that appear on our "Releases" page?

PR #527, opened 2025-03-31
JBR-8408 Post accessibility value changed events for scroll bars

dmitrii-drobotov:
Third-party apps might want to subscribe for scroll bar value changed events to track scroll position. VoiceOver and Zoom don't react to these events. 

https://github.com/user-attachments/assets/36b3ee4b-75c7-447b-8c66-fa6e1e40cc7a


OnePatchGuy (approved):
Hi, thanks for the patch! LGTM

PR #529, opened 2025-03-31
JBR-7700 bugfixes

vladimirlagunov:
Some bug fixes that were made after running performance tests and other non-default tests in IntelliJ.


vladimirlagunov:
The commit `11304b5` should _not_ be cherry-picked to `jbr21`.

PR #557, opened 2025-10-06
JBR-5672 Wayland: support input methods

OnePatchGuy:
# Introduction
This patch allows Java applications running under pure Wayland (`-Dawt.toolkit.name=WLToolkit`) to use input method capabilities provided by the desktop. It's done via utilizing [text-input-unstable-v3](https://wayland.app/protocols/text-input-unstable-v3) protocol.

## Demonstration
[JBR-5672_demo.webm](https://github.com/user-attachments/assets/ca44dd20-ce22-434b-9180-975d72de0a89)

## Build and run
No new build dependencies are required as well as nothing additional is required to make Java apps utilize this feature (though don't forget about `-Dawt.toolkit.name=WLToolkit`. Additionally, the patch provides a new VM option `sun.awt.wl.im.enabled` (it's `true` by default) allowing to disable the feature completely: `-Dsun.awt.wl.im.enabled=false`.

## Important unresolved issues:
- [ ] [zwp_text_input_v3::set_surrounding_text](https://wayland.app/protocols/text-input-unstable-v3#zwp_text_input_v3:request:set_surrounding_text) - [zwp_text_input_v3::delete_surrounding_text](https://wayland.app/protocols/text-input-unstable-v3#zwp_text_input_v3:event:delete_surrounding_text) API isn't supported yet, _**the work is in progress**_. It's an optional API that lets the input method engines to remove/replace already committed text. Besides it's not supported yet, many input method engines work fine on Gnome/Mutter (42.9), e.g. the following of _iBus_ (package names): `ibus-anthy` (Japanese), `ibus-hangul` (Korean), `ibus-libpinyin` (Chinese), `ibus-mozc` (Japanese). Also, _fcitx5_'s pinyin has been checked on KDE 5.27 and it works fine too. **_I'll try to finish work on this by the end of the week, but if I fail, I propose to put off this feature and merge the patch without it to meet the 2025.3 release schedule._**
- [x] When a preedit or commit text gets dispatched to IntelliJ's Editor component, it handles it fine but throws an exception like below. This is a problem on the IntelliJ side ~~, I'm working on the fix, it should land in a Nightly soon.~~ UPD: I couldn't find a _reliable_ solution for this problem, so have introduced a workaround for it in `0fbd1e3ed2f3b324a893df8b3dfa8eb14cc25b02`. Check https://youtrack.jetbrains.com/issue/IJPL-212367 for more info.

<details>
  <summary>Stacktrace</summary>

  ```java
  2025-10-03 17:48:30,282 [  33349] SEVERE - #c.i.o.a.TransactionGuardImpl - Write-unsafe context! Model changes are allowed from write-safe contexts only. Please ensure you're using invokeLater/invokeAndWait with a correct modality state (not "any"). See TransactionGuard documentation for details.
  current modality=ModalityState.NON_MODAL
  java.lang.Throwable: Write-unsafe context! Model changes are allowed from write-safe contexts only. Please ensure you're using invokeLater/invokeAndWait with a correct modality state (not "any"). See TransactionGuard documentation for details.
  current modality=ModalityState.NON_MODAL
	at com.intellij.openapi.diagnostic.Logger.error(Logger.java:375)
	at com.intellij.openapi.application.TransactionGuardImpl.assertWriteActionAllowed(TransactionGuardImpl.java:148)
	at com.intellij.openapi.editor.impl.DocumentImpl.assertWriteAccess(DocumentImpl.java:734)
	at com.intellij.openapi.editor.impl.DocumentImpl.deleteString(DocumentImpl.java:604)
	at com.intellij.openapi.editor.impl.EditorImpl$MyInputMethodHandler.lambda$replaceInputMethodText$3(EditorImpl.java:4310)
	at com.intellij.openapi.application.impl.AppImplKt$runnableUnitFunction$1.invoke(appImpl.kt:104)
	at com.intellij.openapi.application.impl.AppImplKt$runnableUnitFunction$1.invoke(appImpl.kt:104)
	at com.intellij.platform.locking.impl.NestedLocksThreadingSupport.runWriteAction(NestedLocksThreadingSupport.kt:923)
	at com.intellij.openapi.application.impl.ApplicationImpl.runWriteAction(ApplicationImpl.java:1106)
	at com.intellij.openapi.editor.impl.EditorImpl$MyInputMethodHandler.lambda$runUndoTransparent$0(EditorImpl.java:4257)
	at com.intellij.openapi.command.impl.CoreCommandProcessor.lambda$executeCommand$2(CoreCommandProcessor.java:232)
	at com.intellij.openapi.application.impl.AppImplKt$rethrowCheckedExceptions$2.invoke(appImpl.kt:106)
	at com.intellij.platform.locking.impl.NestedLocksThreadingSupport.doRunWriteIntentReadAction(NestedLocksThreadingSupport.kt:666)
	at com.intellij.platform.locking.impl.NestedLocksThreadingSupport.runPreventiveWriteIntentReadAction(NestedLocksThreadingSupport.kt:640)
	at com.intellij.platform.locking.impl.NestedLocksThreadingSupport.runWriteIntentReadAction(NestedLocksThreadingSupport.kt:633)
	at com.intellij.openapi.application.impl.ApplicationImpl.runWriteIntentReadAction(ApplicationImpl.java:1166)
	at com.intellij.openapi.command.impl.CoreCommandProcessor.executeCommand(CoreCommandProcessor.java:228)
	at com.intellij.openapi.command.impl.CoreCommandProcessor.executeCommand(CoreCommandProcessor.java:183)
	at com.intellij.openapi.editor.impl.EditorImpl$MyInputMethodHandler.lambda$runUndoTransparent$1(EditorImpl.java:4257)
	at com.intellij.openapi.command.impl.CoreCommandProcessor.runUndoTransparentAction(CoreCommandProcessor.java:390)
	at com.intellij.openapi.editor.impl.EditorImpl$MyInputMethodHandler.runUndoTransparent(EditorImpl.java:4256)
	at com.intellij.openapi.editor.impl.EditorImpl$MyInputMethodHandler.replaceInputMethodText(EditorImpl.java:4308)
	at com.intellij.openapi.editor.impl.EditorImpl$MyInputMethodListener.inputMethodTextChanged(EditorImpl.java:4077)
	at com.intellij.openapi.editor.impl.EditorComponentImpl.lambda$processInputMethodEvent$0(EditorComponentImpl.java:239)
	at com.intellij.openapi.application.WriteIntentReadAction.lambda$run$0(WriteIntentReadAction.java:24)
	at com.intellij.openapi.application.impl.AppImplKt$rethrowCheckedExceptions$2.invoke(appImpl.kt:106)
	at com.intellij.platform.locking.impl.NestedLocksThreadingSupport.doRunWriteIntentReadAction(NestedLocksThreadingSupport.kt:666)
	at com.intellij.platform.locking.impl.NestedLocksThreadingSupport.runPreventiveWriteIntentReadAction(NestedLocksThreadingSupport.kt:640)
	at com.intellij.platform.locking.impl.NestedLocksThreadingSupport.runWriteIntentReadAction(NestedLocksThreadingSupport.kt:633)
	at com.intellij.openapi.application.impl.ApplicationImpl.runWriteIntentReadAction(ApplicationImpl.java:1166)
	at com.intellij.openapi.application.WriteIntentReadAction.compute(WriteIntentReadAction.java:55)
	at com.intellij.openapi.application.WriteIntentReadAction.run(WriteIntentReadAction.java:23)
	at com.intellij.openapi.editor.impl.EditorComponentImpl.processInputMethodEvent(EditorComponentImpl.java:225)
	at java.desktop/java.awt.Component.processEvent(Component.java:6461)
	at java.desktop/java.awt.Container.processEvent(Container.java:2266)
	at java.desktop/java.awt.Component.dispatchEventImpl(Component.java:5049)
	at java.desktop/java.awt.Container.dispatchEventImpl(Container.java:2324)
	at java.desktop/java.awt.Component.dispatchEvent(Component.java:4877)
	at java.desktop/sun.awt.im.InputMethodContext.dispatchInputMethodEvent(InputMethodContext.java:121)
	at java.desktop/sun.awt.wl.im.text_input_unstable_v3.WLInputMethodZwpTextInputV3.awtDispatchIMESafely(WLInputMethodZwpTextInputV3.java:692)
    ...
  ```

</details>

- [x] When scrolling inside a component while an iBus popup is shown, it's positioned a bit incorrectly (you can observe that in the attached video). It's a bug on Mutter or iBus side - sometimes the IM popup just doesn't react accordingly to [zwp_text_input_v3::set_cursor_rectangle](https://wayland.app/protocols/text-input-unstable-v3#zwp_text_input_v3:request:set_cursor_rectangle) + [zwp_text_input_v3::commit](https://wayland.app/protocols/text-input-unstable-v3#zwp_text_input_v3:request:commit) sent by us. Not sure though they know or have already fixed it, will file a new issue to iBus. 

# Review
I recommend to review the PR commit-by-commit - they have been intentionally formed as "atomic" changes on a single topic, which is described in the commit message. This way I believe it'll be much easier and faster to get to understand the general architecture of the code and how it works in general, as well as how specific routines/mechanics work in details.

### DIFF
--- a/src/java.desktop/unix/classes/sun/awt/wl/im/text_input_unstable_v3/WLInputMethodZwpTextInputV3.java
+++ b/src/java.desktop/unix/classes/sun/awt/wl/im/text_input_unstable_v3/WLInputMethodZwpTextInputV3.java
@@ -26,6 +26,7 @@
 package sun.awt.wl.im.text_input_unstable_v3;
 
 import sun.awt.AWTAccessor;
+import sun.awt.SunToolkit;
 import sun.awt.im.InputMethodAdapter;
 import sun.awt.wl.WLComponentPeer;
 import sun.util.logging.PlatformLogger;
@@ -205,7 +206,7 @@ final class WLInputMethodZwpTextInputV3 extends InputMethodAdapter {
 
         if (getClientComponent() != getPreviousClientComponent()) {
             // The if is required to make sure we don't accidentally lose a still actual preedit string
-            this.awtCurrentClientLatestDispatchedPreeditString = null;
+            this.awtCurrentClientLatestPostedPreeditString = null;
         }
 
         // It may be wrong to invoke this only if awtActivationStatus was DEACTIVATED.
@@ -270,7 +271,7 @@ final class WLInputMethodZwpTextInputV3 extends InputMethodAdapter {
         }
 
         // Deleting the current preedit text
-        awtDispatchIMESafely(JavaPreeditString.EMPTY, JavaCommitString.EMPTY);
+        awtPostIMESafely(JavaPreeditString.EMPTY, JavaCommitString.EMPTY);
 
         // Ending the composition on the native IM's side: the protocol doesn't provide a separate method for this,
         //   but we can achieve this by disabling + enabling back the native context.
@@ -289,7 +290,7 @@ final class WLInputMethodZwpTextInputV3 extends InputMethodAdapter {
 
         awtActivationStatus = AWTActivationStatus.DEACTIVATED;
         awtNativeImIsExplicitlyDisabled = false;
-        awtCurrentClientLatestDispatchedPreeditString = null;
+        awtCurrentClientLatestPostedPreeditString = null;
         awtPreviousClientComponent = null;
         awtClientComponentCaretPositionTracker.stopTrackingCurrentComponent();
         wlDisposeContext();
@@ -309,7 +310,7 @@ final class WLInputMethodZwpTextInputV3 extends InputMethodAdapter {
         sb.append("WLInputMethodZwpTextInputV3@").append(System.identityHashCode(this));
         sb.append('{');
         sb.append("awtActivationStatus=").append(awtActivationStatus);
-        sb.append(", awtCurrentClientLatestDispatchedPreeditString=").append(awtCurrentClientLatestDispatchedPreeditString);
+        sb.append(", awtCurrentClientLatestPostedPreeditString=").append(awtCurrentClientLatestPostedPreeditString);
         sb.append(", wlInputContextState=").append(wlInputContextState);
         sb.append(", wlPendingChanges=").append(wlPendingChanges);
         sb.append(", wlBeingCommittedChanges=").append(wlBeingCommittedChanges);
@@ -357,10 +358,10 @@ final class WLInputMethodZwpTextInputV3 extends InputMethodAdapter {
     /** {@link #setInputMethodContext(InputMethodContext)} */
     private InputMethodContext awtImContext = null;
     /**
-     * {@link #awtDispatchIMESafely(JavaPreeditString, JavaCommitString)}.
-     * {@code null} if no preedit strings have been dispatched since latest {@link #activate}.
+     * {@link #awtPostIMESafely(JavaPreeditString, JavaCommitString)}.
+     * {@code null} if no preedit strings have been posted since latest {@link #activate}.
      */
-    private JavaPreeditString awtCurrentClientLatestDispatchedPreeditString = null;
+    private JavaPreeditString awtCurrentClientLatestPostedPreeditString = null;
     /**
      * Holds the value of {@link #getClientComponent()} since the latest {@link #deactivate(boolean)}
      * It allows to determine whether the focused component has actually changed between the latest deactivate-activate.
@@ -546,10 +547,14 @@ final class WLInputMethodZwpTextInputV3 extends InputMethodAdapter {
     }
 
 
-    /** @return {@code true} if a new InputMethodEvent has been successfully made and dispatched, {@code false} otherwise. */
-    private boolean awtDispatchIMESafely(JavaPreeditString preeditString, JavaCommitString commitString) {
+    // Although we wouldn't break any AWT/Swing rules,
+    //   we can't dispatch InputMethodEvents synchronously instead of posting them to the event queue.
+    // This is because IntelliJ can't properly handle InputEvents that go around the event queue directly to a component.
+    // See TODO: link to the ticket for more info.
+    /** @return {@code true} if a new InputMethodEvent has been successfully made and posted, {@code false} otherwise. */
+    private boolean awtPostIMESafely(JavaPreeditString preeditString, JavaCommitString commitString) {
         if (log.isLoggable(PlatformLogger.Level.FINER)) {
-            log.finer("awtDispatchIMESafely(preeditString={0}, commitString={1}): this={2}.", preeditString, commitString, this);
+            log.finer("awtPostIMESafely(preeditString={0}, commitString={1}): this={2}.", preeditString, commitString, this);
         }
 
         assert(preeditString != null);
@@ -560,7 +565,7 @@ final class WLInputMethodZwpTextInputV3 extends InputMethodAdapter {
                 // Supposing an input method shouldn't interact with UI when not activated.
 
                 if (log.isLoggable(PlatformLogger.Level.FINE)) {
-                    log.fine("awtDispatchIMESafely(preeditString={0}, commitString={1}): ignoring an attempt to dispatch a new InputMethodEvent from an inactive InputMethod={2}.",
+                    log.fine("awtPostIMESafely(preeditString={0}, commitString={1}): ignoring an attempt to post a new InputMethodEvent from an inactive InputMethod={2}.",
                              preeditString, commitString, this);
                 }
 
@@ -573,7 +578,7 @@ final class WLInputMethodZwpTextInputV3 extends InputMethodAdapter {
 
                 if (log.isLoggable(PlatformLogger.Level.WARNING)) {
                     log.warning(
-                        "awtDispatchIMESafely(preeditString={0}, commitString={1}): can''t dispatch a new InputMethodEvent because there''s no client component to dispatch to, although this InputMethod is active. this={2}.",
+                        "awtPostIMESafely(preeditString={0}, commitString={1}): can''t post a new InputMethodEvent because there''s no client component to dispatch to, although this InputMethod is active. this={2}.",
                         preeditString, commitString, this
                     );
                 }
@@ -607,7 +612,7 @@ final class WLInputMethodZwpTextInputV3 extends InputMethodAdapter {
                 }
             }
 
-            final var clientCurrentPreeditText = Objects.requireNonNullElse(this.awtCurrentClientLatestDispatchedPreeditString, JavaPreeditString.EMPTY);
+            final var clientCurrentPreeditText = Objects.requireNonNullElse(this.awtCurrentClientLatestPostedPreeditString, JavaPreeditString.EMPTY);
             if ( commitString.equals(JavaCommitString.EMPTY) &&
                  preeditString.equals(JavaPreeditString.EMPTY) &&
                  clientCurrentPreeditText.equals(JavaPreeditString.EMPTY) )
@@ -623,8 +628,8 @@ final class WLInputMethodZwpTextInputV3 extends InputMethodAdapter {
                 //     text attributes and god knows what else.
 
                 if (log.isLoggable(PlatformLogger.Level.FINE)) {
-                    log.fine("awtDispatchIMESafely(preeditString={0}, commitString={1}): ignoring a redundant attempt to dispatch a new empty InputMethodEvent to a component with no preedit text. this={2}.",
-                            preeditString, commitString, this);
+                    log.fine("awtPostIMESafely(preeditString={0}, commitString={1}): ignoring a redundant attempt to post a new empty InputMethodEvent to a component with no preedit text. this={2}.",
+                             preeditString, commitString, this);
                 }
 
                 return false;
@@ -679,24 +684,8 @@ final class WLInputMethodZwpTextInputV3 extends InputMethodAdapter {
                 }
             }
 
-            if (log.isLoggable(PlatformLogger.Level.FINEST)) {
-                log.finest(
-                    "awtDispatchIMESafely(...): dispatching an InputMethodEvent equal to {0}. this={1}.",
-                    // the parameters of this fictitious IME must be synchronized with what's actually dispatched below
-                    new InputMethodEvent(
-                        clientComponent,
-                        InputMethodEvent.INPUT_METHOD_TEXT_CHANGED,
-                        imeText.getIterator(),
-                        commitString.text().length(),
-                        imeCaret,
-                        null
-                    ),
-                    this
-                );
-            }
-
-            // NB: keep the parameters passed here synchronized with what's logged above.
-            this.awtImContext.dispatchInputMethodEvent(
+            final InputMethodEvent ime = new InputMethodEvent(
+                clientComponent,
                 InputMethodEvent.INPUT_METHOD_TEXT_CHANGED,
                 imeText.getIterator(),
                 commitString.text().length(),
@@ -704,11 +693,20 @@ final class WLInputMethodZwpTextInputV3 extends InputMethodAdapter {
                 // no recommendation for a visible position within the current preedit text
                 null
             );
-            this.awtCurrentClientLatestDispatchedPreeditString = preeditString;
+
+            if (log.isLoggable(PlatformLogger.Level.FINEST)) {
+                log.finest(
+                    String.format("awtPostIMESafely(...): posting a new InputMethodEvent=%s. this=%s.", ime, this),
+                    new Throwable("Stacktrace")
+                );
+            }
+
+            SunToolkit.postEvent(SunToolkit.targetToAppContext(clientComponent), ime);
+            this.awtCurrentClientLatestPostedPreeditString = preeditString;
 
             return true;
         } catch (Exception err) {
-            log.severe("Error occurred during constructing and dispatching a new InputMethodEvent.", err);
+            log.severe("Error occurred during constructing or posting a new InputMethodEvent.", err);
         }
 
         return false;
@@ -1130,10 +1128,10 @@ final class WLInputMethodZwpTextInputV3 extends InputMethodAdapter {
             // Erasing preedit text from the client component
             if (awtActivationStatus == AWTActivationStatus.ACTIVATED) {
                 final boolean preeditIsEmpty =
-                    (awtCurrentClientLatestDispatchedPreeditString == null)
-                    || awtCurrentClientLatestDispatchedPreeditString.text().isEmpty();
+                    (awtCurrentClientLatestPostedPreeditString == null)
+                    || awtCurrentClientLatestPostedPreeditString.text().isEmpty();
                 if (!preeditIsEmpty) {
-                    awtDispatchIMESafely(JavaPreeditString.EMPTY, JavaCommitString.EMPTY);
+                    awtPostIMESafely(JavaPreeditString.EMPTY, JavaCommitString.EMPTY);
                 }
             }
         } catch (Exception err) {
@@ -1437,7 +1435,7 @@ final class WLInputMethodZwpTextInputV3 extends InputMethodAdapter {
             //   +-------+----------------------------------------+          |
             //          1|                                                   |
             //           v                                           +-------+------------------------------------+
-            //    awtDispatchIMESafely(...)                          |zwp_text_input_v3::set_cursor_rectangle(...)|
+            //    awtPostIMESafely(...)                              |zwp_text_input_v3::set_cursor_rectangle(...)|
             //          2|                                           |                                            |
             //           v                                           |zwp_text_input_v3::commit()                 |
             //    the component deletes its current composed text    +-------^------------------------------------+
@@ -1479,12 +1477,14 @@ final class WLInputMethodZwpTextInputV3 extends InputMethodAdapter {
             //
             // TODO: unconditional dispatching of the text makes it impossible for the caret to leave the component's
             //       visible area, which would be useful for scrolling.
-            awtDispatchIMESafely(preeditStringToApply, commitStringToApply);
+            awtPostIMESafely(preeditStringToApply, commitStringToApply);
 
             if (doDeferCaretPositionUpdates) {
                 // invokeLater is needed to handle a case when the client component decides to post the caret updates
                 //   to the EventQueue instead of dispatching them right away (IDK if it happens in practice).
-                EventQueue.invokeLater(() -> {
+                // Nested invokeLater is needed because the IME event hasn't been dispatched yet (only posted).
+                // This way we guarantee that all UI activities associated with it are finished before resuming.
+                EventQueue.invokeLater(() -> EventQueue.invokeLater(() -> {
                     try {
                         this.awtClientComponentCaretPositionTracker.resumeUpdates(false);
                     } catch (Exception err) {
@@ -1494,7 +1494,7 @@ final class WLInputMethodZwpTextInputV3 extends InputMethodAdapter {
                             err
                         );
                     }
-                });
+                }));
             }
 
             // Sending pending changes (if any)


src/java.desktop/unix/classes/sun/awt/wl/im/text_input_unstable_v3/WLInputMethodZwpTextInputV3.java
```diff
@@ -0,0 +1,191 @@
...
+
+import sun.awt.im.InputMethodAdapter;
+import sun.util.logging.PlatformLogger;
+
+import java.awt.*;
+import java.awt.im.spi.InputMethodContext;
+import java.util.Locale;
+
+
+final class WLInputMethodZwpTextInputV3 extends InputMethodAdapter {
```
OnePatchGuy, line 53: The package name + the class name together may sound a little over-complicated or like putting a hat on a hat: _text_input_unstable_v3_ in the package name and the _ZwpTextInputV3_ postfix in the class name, but the latter has been done on purpose: I've noticed that InteliiJ sometimes truncates or obfuscates full names of the loggers, so that only the name of the class is fully visible. If we implement several IM protocols, it may not be obvious which one is being used if we don't add such postfix.

src/java.desktop/unix/classes/sun/awt/wl/im/text_input_unstable_v3/WLInputMethodZwpTextInputV3.java
```diff
@@ -186,7 +202,14 @@ public void deactivate(boolean isTemporary) {
...
     public void hideWindows() {
-        // TODO: implement
+        // "The method is only called when the input method is inactive."
+        assert(this.awtActivationStatus != AWTActivationStatus.ACTIVATED);
+
+        // The protocol doesn't provide a separate method for hiding the IM window(s),
+        //   but this effect can be achieved by disabling the native context.
+        // Actually, it should have already been disabled (because the IM is inactive),
+        //   so disabling here is performed just in case.
+        wlDisableContextNow();
```
OnePatchGuy, line 252: Initially, the procedure was `wlDisableContextNow()` + `wlEnableContextNow()` that would just reset the state of the native IM and hide its popup until the user starts typing again. But then I read in this method's documentation that it can only be called when the instance is inactive, hence it just doesn't make sense to try to enable it. This is how only `wlDisableContextNow()` is left.

src/java.desktop/unix/classes/sun/awt/wl/im/text_input_unstable_v3/WLInputMethodZwpTextInputV3.java
```diff
@@ -885,8 +885,6 @@ private void wlSendPendingChangesNow() {
             return;
         }
 
-        // TODO: make sure the current AWT component belongs to the surface wlInputContextState.getCurrentWlSurfacePtr()
```
OnePatchGuy: Well, even if we detected that the current AWT component belongs to another wl_surface, what a good response would be in that case? I don't think it's a good idea to blow up everything with exceptions and stop working, but the only other thing that comes to my head is just to log such a case and move on, which seems anyway useless. So I've just dropped this TODO

src/java.desktop/unix/classes/sun/awt/wl/WLToolkit.java
```diff
@@ -882,12 +882,6 @@ public InputMethodDescriptor getInputMethodAdapterDescriptor() {
      */
     @Override
     public boolean enableInputMethodsForTextComponent() {
-        if (log.isLoggable(PlatformLogger.Level.FINE)) {
-            log.fine("Not implemented: WLToolkit.enableInputMethodsForTextComponent()");
-        }
-
-        // TODO: implement
```
OnePatchGuy: Already implemented :)

mkartashev (commented):
Running jbr21 with this change and `-Xcheck:jni` produces several warnings that need to be addressed before merging. There seem to be two distinct ones:

* `Warning: Calling other JNI functions in the scope of Get/ReleasePrimitiveArrayCritical or Get/ReleaseStringCritical`
This one reproduces even with `-Dsun.awt.wl.im.enabled=false`, but no such warning without this patch (although I compared with `main`).

* Another one is somewhere in the native event handling code with the stack trace like this:
```
WARNING in native method: JNI call made without checking exceptions when required to from CallVoidMethod
	at sun.awt.wl.WLToolkit.dispatchEventsOnEDT(java.desktop@21.0.8-internal/Native Method)
	at sun.awt.wl.WLToolkit.lambda$run$2(java.desktop@21.0.8-internal/WLToolkit.java:261)
	at sun.awt.wl.WLToolkit$$Lambda/0x00007e2470072508.run(java.desktop@21.0.8-internal/Unknown Source)
	at java.awt.event.InvocationEvent.dispatch(java.desktop@21.0.8-internal/InvocationEvent.java:318)
	at java.awt.EventQueue.dispatchEventImpl(java.desktop@21.0.8-internal/EventQueue.java:781)
	at java.awt.EventQueue$4.run(java.desktop@21.0.8-internal/EventQueue.java:728)
	at java.awt.EventQueue$4.run(java.desktop@21.0.8-internal/EventQueue.java:722)
	at java.security.AccessController.executePrivileged(java.base@21.0.8-internal/AccessController.java:778)
	at java.security.AccessController.doPrivileged(java.base@21.0.8-internal/AccessController.java:400)
	at java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(java.base@21.0.8-internal/ProtectionDomain.java:87)
	at java.awt.EventQueue.dispatchEvent(java.desktop@21.0.8-internal/EventQueue.java:750)
	at java.awt.EventDispatchThread.pumpOneEventForFilters(java.desktop@21.0.8-internal/EventDispatchThread.java:207)
	at java.awt.EventDispatchThread.pumpEventsForFilter(java.desktop@21.0.8-internal/EventDispatchThread.java:128)
	at java.awt.EventDispatchThread.pumpEventsForHierarchy(java.desktop@21.0.8-internal/EventDispatchThread.java:117)
	at java.awt.EventDispatchThread.pumpEvents(java.desktop@21.0.8-internal/EventDispatchThread.java:113)
	at java.awt.EventDispatchThread.pumpEvents(java.desktop@21.0.8-internal/EventDispatchThread.java:105)
	at java.awt.EventDispatchThread.run(java.desktop@21.0.8-internal/EventDispatchThread.java:92)
```
It's hard to pinpoint the culprit from the stack trace. If I spot a suspicious code during the review, I will communicate separately. This warning disappears with the introduction of the `-Dsun.awt.wl.im.enabled=false` option.

They can be reproduced simply by running 
`./build/linux-x86_64-server-release/images/jdk/bin/java -Dawt.toolkit.name=WLToolkit  -Xcheck:jni -jar ./build/linux-x86_64-server-release/images/jdk/demo/jfc/Stylepad/Stylepad.jar`

src/java.desktop/unix/native/libawt_wlawt/im/text-input-unstable-v3/WLInputMethod.c
```diff
@@ -0,0 +1,668 @@
...
+    if (imContext == NULL) {
+        return;
+    }
+
+    env = getEnv();
+    if (env == NULL) {
+        return;
+    }
+
+    (*env)->CallVoidMethod(env, imContext->wlInputMethodOwner, jniIDs.mID_tiOnEnter, ptr_to_jlong(surface));
```
mkartashev, line 221: Need to `JNU_CHECK_EXCEPTION` (or something similar) in all instances like this.

src/java.desktop/unix/native/libawt_wlawt/im/text-input-unstable-v3/WLInputMethod.c
```diff
@@ -0,0 +1,668 @@
...
+
+// The IMContext_zwp_text_input_v3_on* callbacks are supposed to be as a thin bridge to
+// `WLInputMethodZwpTextInputV3`'s JNI upcalls as possible in terms of contained logic.
+// Generally they should only invoke the corresponding upcalls with the received parameters.
+
+static void IMContext_zwp_text_input_v3_onEnter(
+    void * const ctx,
+    struct zwp_text_input_v3 * const textInput,
+    struct wl_surface * const surface
+) {
```
mkartashev, line 208: Formatting of function declarators is somewhat inconsistent in this file. There's only a style guide for HotSpot code and it suggest to have `{` on the same line as the function declarator like it is here. But then you will have to change all the other declarations, so I don't insist on one specific formatting style. Just make it consistent.

mkartashev (changes requested):

src/java.desktop/unix/classes/sun/awt/wl/im/WLInputMethodMetaDescriptor.java
```diff
@@ -0,0 +1,97 @@
...
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+package sun.awt.wl.im;
+
+import sun.util.logging.PlatformLogger;
+
+import java.awt.*;
```
mkartashev: Wildcard imports are discouraged in the JDK code.

mkartashev:
> After reverting all PR's commits and making a clean JBR build, the warnings are still here, hence I consider this not a problem with the patch.

Thanks for checking! I will file a separate issue for this.

src/java.desktop/unix/native/libawt_wlawt/im/text-input-unstable-v3/WLInputMethod.c
OnePatchGuy: Thanks! Fixed in `956451ec891388591dcda8d27f6ea50bfc5481db`. 

It also resolves the warnings you were talking about [above](https://github.com/JetBrains/JetBrainsRuntime/pull/557#pullrequestreview-3309103345):
> WARNING in native method: JNI call made without checking exceptions when required to from CallVoidMethod

However, if you try to input in a simple JTextArea using an input method (not just regular keyboard input), you'll still be getting tons of the following warnings under `-Xcheck:jni`:
> WARNING in native method: JNI call made without checking exceptions when required to from CallObjectMethodV

But they all come from an OpenJDK's `src/java.desktop/share/native/libfontmanager/hb-jdk-font.cc`, where `CallObjectMethod`s aren't followed by `ExceptionCheck`s. It needs to be filed to OpenJDK.

src/java.desktop/unix/native/libawt_wlawt/im/text-input-unstable-v3/WLInputMethod.c
OnePatchGuy: Fair note, thanks! Forgot to put OpenJDK's codestyle over my personal preferences while developing.
If you don't mind, I'd stop at the following rules regarding functions signatures here:
* If it fits one line, it should be `<return-type-etc> <function-name>(<parameters-list>) {`
* If `<parameters-list>` is too long, the signature will be split into lines like that:
```
<return-type-etc> <function-name>(
    <parameter-1>,
    ...,
    <parameter-N>
) {
```
* In case of JNI downcalls definitions, `<return-type-etc>` is put on a separate line (because `JNIEXPORT <return-type> JNICALL` may take quite a lot of space)

Made corresponding adjustments in `04936ea9d64e79b0828b0a14cf67906b040ad1a8`,

src/java.desktop/unix/classes/sun/awt/wl/im/text_input_unstable_v3/InputContextState.java
```diff
@@ -0,0 +1,114 @@
...
+    /** {@code null} if the InputContextState is disabled. */
+    private StateOfEnabled stateOfEnabled = null;
+    /**
+     * The latest preedit string applied as a result of the latest {@code zwp_text_input_v3::done} event received.
+     * Must never be {@code null} ; if a {@code zwp_text_input_v3::done} event wasn't preceded by a
+     * {@code zwp_text_input_v3::preedit_string} event, the field should be set to {@link PropertiesInitials#PREEDIT_STRING}.
+     */
+    private JavaPreeditString latestAppliedPreeditString = PropertiesInitials.PREEDIT_STRING;
+    /**
+     * The latest preedit string applied as a result of the latest {@code zwp_text_input_v3::done} event received.
```
mkartashev: Is this line of the comment accurate? It looks identical to the description of the previous field `latestAppliedPreeditString`.

src/java.desktop/unix/classes/sun/awt/wl/im/text_input_unstable_v3/WLInputMethodZwpTextInputV3.java
```diff
@@ -174,6 +179,13 @@ public Object getControlObject() {
 
     /* java.lang.Object methods section */
 
+    @SuppressWarnings("removal")
+    @Override
+    protected void finalize() throws Throwable {
+        dispose();
```
mkartashev: Not sure the finalize needed here; moreover, it doesn't seem to be safe as under normal circumstances `dispose()` would've been called prior to `finalize()`. And yet we call `dispose()` here again breaking the contract of `InputMethod.dispose()`:
```
No method of this interface is called on this instance after dispose.
```

src/java.desktop/unix/classes/sun/awt/wl/WLComponentPeer.java
```diff
@@ -1590,7 +1590,7 @@ int surfaceUnitsToJavaSize(int value) {
      * Converts a value in the Java coordinate system into the Wayland
      * surface-local coordinate system.
      */
-    int javaUnitsToSurfaceUnits(int value) {
+    public int javaUnitsToSurfaceUnits(int value) {
```
mkartashev: Probably makes sense to make them `final` now that they are used even outside of this package.
Should've done it myself earlier, of course.

src/java.desktop/unix/classes/sun/awt/wl/im/WLInputMethodMetaDescriptor.java
```diff
@@ -180,6 +184,25 @@ public String toString() {
...
+     *   unnecessary for them functionality if they face any problems because of it.
+     * Therefore, if it's {@code false}, the Toolkit code shouldn't use (directly or indirectly)
+     *   any of Wayland's input methods-related APIs (e.g. the "text-input" protocol).
+     */
+    private final static boolean ENABLE_NATIVE_IM_SUPPORT;
+
+    static {
+        boolean enableNativeImSupportInitializer = true;
+        try {
+            enableNativeImSupportInitializer = Boolean.parseBoolean(System.getProperty("sun.awt.wl.im.enabled", "true"));
```
mkartashev: In jbr21 you need to wrap these things into `AccessController.doPrivileged()` or else tests might start to complain.

src/java.desktop/unix/native/libawt_wlawt/im/text-input-unstable-v3/WLInputMethod.c
```diff
@@ -0,0 +1,672 @@
...
+Java_sun_awt_wl_im_text_1input_1unstable_1v3_WLInputMethodZwpTextInputV3_initIDs(JNIEnv * const env, const jclass clazz) {
+    CHECK_NULL_THROW_OOME(
+        env,
+        jniIDs.wlInputMethodClass = (*env)->NewGlobalRef(env, clazz),
+        "Allocation of a global reference to sun.awt.wl.im.text_input_unstable_v3.WLInputMethodZwpTextInputV3 class failed"
+    );
+
+    jniIDs.mID_tiOnEnter =
+        (*env)->GetMethodID(env, jniIDs.wlInputMethodClass, "zwp_text_input_v3_onEnter", "(J)V");
+    if (jniIDs.mID_tiOnEnter == NULL)
```
tsarn: style: Inconsistent brace style compared to most other if statements in this file

src/java.desktop/unix/native/libawt_wlawt/im/text-input-unstable-v3/WLInputMethod.c
```diff
@@ -0,0 +1,672 @@
...
+        return;
+    }
+
+    env = getEnv();
+    if (env == NULL) {
+        return;
+    }
+
+    if (commitStringUtf8 != NULL) {
+        // Theoretically, a UTF-8 string may:
```
tsarn: Maybe move this code to a common function, since it's shared with `IMContext_zwp_text_input_v3_onPreeditString`?

src/java.desktop/unix/native/libawt_wlawt/im/text-input-unstable-v3/WLInputMethod.c
```diff
@@ -0,0 +1,672 @@
...
+
+    textInput = imContext->textInput;
+    if (textInput == NULL) {
+        JNU_ThrowByName(env, "java/lang/IllegalStateException", "textInput == NULL");
+        return;
+    }
+
+    zwp_text_input_v3_set_cursor_rectangle(
+        textInput,
+        (int32_t)surfaceLocalX,
```
tsarn, line 608: missing `#include <stdint.h>`

src/java.desktop/unix/native/libawt_wlawt/WLToolkit.c
```diff
@@ -610,6 +612,14 @@ registry_global(void *data, struct wl_registry *wl_registry,
             WLOutputXdgOutputManagerBecameAvailable();
             process_new_listener_before_end_of_init();
         }
+    } else if (strcmp(interface, zwp_text_input_manager_v3_interface.name) == 0) {
+        // If the requested version is higher than the provided one by the compositor,
+        //   the event loop may shut down as soon as it gets launched (wl_display_dispatch will return -1),
+        //   so let's protect from this since the component being obtained is not vital for work.
+        const uint32_t versionToBind = (uint32_t)zwp_text_input_manager_v3_interface.version;
```
tsarn: Maybe it's better to bind to a specific version, 1 in this case? Future versions may deprecate some methods or make it an error to call certain methods with certain parameters. See for instance [wl_surface::attach](https://wayland.app/protocols/wayland#wl_surface:request:attach), passing nonzero x and y is a protocol violation for protocol versions >= 5.

src/java.desktop/unix/classes/sun/awt/wl/im/WLInputMethodMetaDescriptor.java
OnePatchGuy: Is that really true? _None_ of the usages of `System.getProperty` inside the `sun.awt.wl` package in jbr21 is wrapped in `doPrivileged()`

src/java.desktop/unix/native/libawt_wlawt/im/text-input-unstable-v3/WLInputMethod.c
OnePatchGuy: I believe these typedefs should come from the wayland headers we include because they're directly used in the generated sources of protocols. If you don't mind I'd just add a comment at the corresponding include.

tsarn: Yeah, it's not important one way or the other, it's just best practice to explicitly include all the names used and not rely on transitive includes.

src/java.desktop/unix/classes/sun/awt/wl/im/WLInputMethodMetaDescriptor.java
OnePatchGuy: Good point! Fixed in `58b911eddf3e89ae0507bc5ad8645f89d497bb73`.

src/java.desktop/unix/classes/sun/awt/wl/im/text_input_unstable_v3/InputContextState.java
OnePatchGuy: What a keen eye, thanks! Fixed in `d14905bb6b6c36d0338de5dbc3928fcb666efe7b`.

src/java.desktop/unix/classes/sun/awt/wl/im/text_input_unstable_v3/WLInputMethodZwpTextInputV3.java
OnePatchGuy: Initially I've written it, you know, just in case, but a formal inability to call `dispose` after `dispose` makes a good point to remove this override, thank you! Fixed in `3eeaef669fa74f14e4205c8026c762bc16ee6c5f`.

src/java.desktop/unix/classes/sun/awt/wl/WLComponentPeer.java
OnePatchGuy: Alrighty, done in `3ddd977ebe7b94bdaf6d47fd06e080be503b14f6`.

src/java.desktop/unix/native/libawt_wlawt/im/text-input-unstable-v3/WLInputMethod.c
OnePatchGuy: Fair, thanks! Fixed in `91882ca140fb9579e285b25611cec042ca7f6987`.

src/java.desktop/unix/native/libawt_wlawt/im/text-input-unstable-v3/WLInputMethod.c
OnePatchGuy: Added a comment in `83685af7b5da469a8a7e092d9d80fa9dd443a3c3`.

src/java.desktop/unix/native/libawt_wlawt/im/text-input-unstable-v3/WLInputMethod.c
OnePatchGuy: A good suggestion, thanks! Extracted in a separate function in `283fa74d5a7e92047999221137b874d07769465f`.

src/java.desktop/unix/native/libawt_wlawt/im/text-input-unstable-v3/WLInputMethod.c
OnePatchGuy: I realized that may have confused you in my first comment, the thing I wanted to say is that in this file _we'd better rely on the version of these typedefs coming from wayland headers rather than from the stdlib_.

src/java.desktop/unix/native/libawt_wlawt/WLToolkit.c
OnePatchGuy: Great catch! I've checked what they're doing for the update of the protocol ([there](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/282)) and yes, the semantics of already existing methods may change although they'll be binary compatible, and we'll have to take that into account.

I'll make a change binding to version `1`, but here's a question to think about: since in _jbr25_ and _main_ we generate the interfaces from xml, once it can happen we're trying to bind using version `1` when the generated `zwp_text_input_manager_v3_interface` is of version `2`, will such usage be safe? E.g. what will happen with an event that exists only in v.2?

The protocol itself promises backward compatibility, but it's not clear what exactly such compatibility means:
> Backward compatible changes may be added together with the corresponding interface version bump. Backward incompatible changes are done by bumping the version number in the protocol and interface names and resetting the interface version.

tsarn: This should be safe, I think the events from newer versions than announced won't be emitted, everything should be binary compatible.

I believe we are already using newer headers than protocol versions for some protocols.

OnePatchGuy: Made the corresponding change in `26904fd5f45cf2670aba636ffc76a8ccdc23cd6d`. With it, having that `if` looks useless, but I'd rather keep it due to the comment above which I find pretty useful.

src/java.desktop/unix/classes/sun/awt/wl/im/WLInputMethodMetaDescriptor.java
mkartashev: Yes, it's true for jbr21, but not for later versions. The access to system properties may not be wrapped directly, but is called in the right context. See for example WLToolkit:
```
        AccessController.doPrivileged((PrivilegedAction<Void>) () -> {
            initSystemProperties();
            return null;
        });
```

OnePatchGuy: Ok, I see. Thanks for the info! I thought all these stuff had been completely disabled for Java 21.

Fixed in `e1359ab2c72d0329c6fa3ac228ce837808321c75`.

mkartashev: Sadly, no, just deprecated in 21.

PR #566, opened 2025-10-21
JBR-9515 Allow size of per-directory buffer used to retrieve events to be configurable to avoid OVERFLOW_EVENT

k15tfu:


src/java.base/windows/classes/sun/nio/fs/WindowsWatchService.java
```diff
@@ -284,9 +285,26 @@ private static class Poller extends AbstractPoller {
...
+                String.valueOf(DEFAULT_CHANGES_BUFFER_SIZE));
+            int intValue;
+            try {
+                // Clamp to size of per-directory buffer used to retrieve events.
+                intValue = Math.clamp(
+                    Long.decode(rawValue),
+                    1,
+                    ArraysSupport.SOFT_MAX_ARRAY_LENGTH);
+            } catch (NumberFormatException e) {
+                intValue = DEFAULT_CHANGES_BUFFER_SIZE;
```
shalupov, line 304: May we add a warning here?

src/java.base/windows/classes/sun/nio/fs/WindowsWatchService.java
```diff
@@ -284,9 +285,26 @@ private static class Poller extends AbstractPoller {
...
+        private static final int DEFAULT_CHANGES_BUFFER_SIZE = 16 * 1024;
+        static final int CHANGES_BUFFER_SIZE;
+        static {
+            String rawValue = System.getProperty(
+                "jdk.nio.file.WatchService.bufferSizeToRetrieveEventsPerDirectory",
+                String.valueOf(DEFAULT_CHANGES_BUFFER_SIZE));
+            int intValue;
+            try {
+                // Clamp to size of per-directory buffer used to retrieve events.
+                intValue = Math.clamp(
```
shalupov, line 299: May we add a warning (stderr?) or logger here if intValue is different from rawValue?

src/java.base/windows/classes/sun/nio/fs/WindowsWatchService.java
mkartashev: What about the comment stating that
> Need to be less than 4*16384 = 65536 when monitoring a directory over the network

? Are the users of these options supposed to be responsible enough to observe the limit? Or is the limit outdated?

k15tfu: > Are the users of these options supposed to be responsible enough to observe the limit?

Otherwise it won’t be possible to set a limit higher than 64K. If it's better, we could add an additional flag to explicitly allow this behavior.

> Or is the limit outdated?

No.

PR #571, opened 2025-10-24
JBR-9531 Prevent unexpected recursive usage of java.io over nio wrappers

vladimirlagunov:
Shortly said, we don't need to handle recursive invocations of the `java.io` to nio adapters, while this recursion leads to problems.

# Problem 

Since we're not allowed to modify public API of Java classes, an unreliable workaround `IoOverNio.PARENT_FOR_FILE_CHANNEL_IMPL` had been introduced. This trick lets us pass some object into a called function without adding a new function argument.

The trick backfired at the following code:

```java
// at java.base/java.io.IoOverNioFileSystem.initializeStreamUsingNio (simplified code)
IoOverNio.PARENT_FOR_FILE_CHANNEL_IMPL.set(owner);
return initializeStreamsUsingNio0(owner, nioFs, file, nioPath, optionsForChannel, channelCleanable);


// at java.base/java.io.IoOverNioFileSystem.initializeStreamsUsingNio0
channel = nioFs.provider().newFileChannel(nioPath, optionsForChannel);
```

The intention of setting `PARENT_FOR_FILE_CHANNEL_IMPL` is to path `owner` inside a constructor of `sun.nio.ch.FileChannelImpl` which may be called by `newFileChannel(nioPath, optionsForChannel)`.

The implementation of `newFileChannel` in IntelliJ triggered class loading. The classloader of IntelliJ triggered the call `java.nio.file.Files.readAllBytes`. The latter code also invoked the constructor of `FileChannelImpl`, and the constructor got from the thread-local variable the value for the field `parent`. That value was supposed to be passed to a different invocation of the constructor.

The observable result of this bug was a `NullPointerExceptions` from `java.io.FileOutputStream.close`. Hypothetically, this bug could also lead to closing file descriptors earlier as expected in other unpredictable places.

# Rejected approaches

* The cleanest solution could be passing specific function arguments through function arguments. 
  
  However, we may neither create a new public function in the module `java.base`, nor add an argument to an existing one. Thus, we have to keep dealing with thread-local variables.
* To hold several values in the thread-local variable. 

  In this case, the constructor of `FileChannelImpl` should somehow choose the right value from the list. The only possible way for filtering values is by the provided path. It doesn't look like a performant and reliable solution.
 
# Chosen solution
 
This commit disables recursive invocations of `java.io` from `java.nio` adapters.
  
The chosen solution is tied to specifics of IntelliJ. The reason for introducing java.io over java.nio adapter was to be able to access files from remote machines without rewriting old code. It is not expected that an implementation of `java.nio.file.spi.FileSystemProvider` accesses the file system using `java.io`. The only imaginable way to get a `java.io` call from `java.nio` is class loading. In case of IntelliJ, it's assumed that the class loader never accesses classes and jars from a remote machine.

### DIFF
--- a/src/java.base/share/classes/java/io/IoOverNioFileSystem.java
+++ b/src/java.base/share/classes/java/io/IoOverNioFileSystem.java
@@ -270,18 +270,13 @@ class IoOverNioFileSystem extends FileSystem {
             Path nioPath,
             Set<OpenOption> optionsForChannel,
             NioChannelCleanable channelCleanable) throws FileNotFoundException {
-        var oldOwner = IoOverNio.ParentForFileChannelImplHolder.get();
         try {
             // This tricky thread local variable allows specifying an argument for sun.nio.ch.FileChannelImpl.<init>
             // which is not present in the NIO public API and which is not easy to specify another way.
             IoOverNio.ParentForFileChannelImplHolder.set(owner);
             return initializeStreamsUsingNio0(owner, nioFs, file, nioPath, optionsForChannel, channelCleanable);
         } finally {
-            if (oldOwner != null) {
-                IoOverNio.ParentForFileChannelImplHolder.set(oldOwner);
-            } else {
-                IoOverNio.ParentForFileChannelImplHolder.remove();
-            }
+            IoOverNio.ParentForFileChannelImplHolder.remove();
         }
     }
 


src/java.base/share/classes/java/io/IoOverNioFileSystem.java
```diff
@@ -270,13 +270,18 @@ static StreamInitializationBundle initializeStreamUsingNio(
...
             NioChannelCleanable channelCleanable) throws FileNotFoundException {
+        var oldOwner = IoOverNio.PARENT_FOR_FILE_CHANNEL_IMPL.get();
         try {
             // This tricky thread local variable allows specifying an argument for sun.nio.ch.FileChannelImpl.<init>
             // which is not present in the NIO public API and which is not easy to specify another way.
             IoOverNio.PARENT_FOR_FILE_CHANNEL_IMPL.set(owner);
             return initializeStreamsUsingNio0(owner, nioFs, file, nioPath, optionsForChannel, channelCleanable);
         } finally {
-            IoOverNio.PARENT_FOR_FILE_CHANNEL_IMPL.remove();
+            if (oldOwner != null) {
```
mkartashev: Could you please clarify this code: there's a provision for recursive use, but the whole idea of the change is to avoid recursive use.

src/java.base/share/classes/java/io/IoOverNioFileSystem.java
```diff
@@ -270,13 +270,18 @@ static StreamInitializationBundle initializeStreamUsingNio(
             Path nioPath,
             Set<OpenOption> optionsForChannel,
             NioChannelCleanable channelCleanable) throws FileNotFoundException {
+        var oldOwner = IoOverNio.PARENT_FOR_FILE_CHANNEL_IMPL.get();
```
mkartashev: Is every use of `IoOverNio.PARENT_FOR_FILE_CHANNEL_IMPL` supposed to be guarded by `RecursionGuard`? If so, it'd be nice to have a check here; if the check is inexpensive, you can even keep it in the release mode as violations of such a protocol are rather expensive to track down.

src/java.base/share/classes/java/io/FileInputStream.java
```diff
@@ -156,37 +156,40 @@ public FileInputStream(File file) throws FileNotFoundException {
...
 
-        java.nio.file.FileSystem nioFs = IoOverNioFileSystem.acquireNioFs(path);
-        Path nioPath = null;
-        if (nioFs != null && path != null) {
-            try {
-                nioPath = nioFs.getPath(path);
-                isRegularFile = Files.isRegularFile(nioPath);
-            } catch (InvalidPathException _) {
-                // Nothing.
+        try (var guard = IoOverNio.RecursionGuard.create(FileInputStream.class)) {
```
mkartashev, line 159: I'm a bit concerned how a potential future new author of a similar code will learn of the necessity to use `RecursionGuard`, especially that the failure to use it introduces subtle bugs?
Can it be enforced at some level with an assert or some other check? Maybe introduce a dummy argument of the type `RecursionGuard` to all the necessary methods?

src/java.base/share/classes/java/io/IoOverNioFileSystem.java
vladimirlagunov: You're right, this code is redundant.

src/java.base/share/classes/java/io/FileInputStream.java
vladimirlagunov: https://github.com/JetBrains/JetBrainsRuntime/pull/571/commits/c613a3bf07554d874488ac4d174b2f0573ac58bc

src/java.base/share/classes/java/io/IoOverNioFileSystem.java
vladimirlagunov: https://github.com/JetBrains/JetBrainsRuntime/pull/571/commits/c613a3bf07554d874488ac4d174b2f0573ac58bc

src/java.base/share/classes/java/io/IoOverNioFileSystem.java
vladimirlagunov: https://github.com/JetBrains/JetBrainsRuntime/pull/571/commits/8a4e2b45b2fbadbf608be3ff492f5aa31b2e3962

mkartashev (approved):

PR #587, opened 2026-02-11
JBR-9994 Enable AccessibleAnnouncer for both X11 and Wayland

tanya011:
Move AccessibleAnnouncer to shared libawt location and added speech-dispatcher configuration to libawt_wlawt build.


make/modules/java.desktop/lib/Awt2dLibraries.gmk
```diff
@@ -204,7 +212,7 @@ $(eval $(call SetupJdkLibrary, BUILD_LIBAWT, \
         -delayload:ole32.dll -delayload:comdlg32.dll \
         -delayload:comctl32.dll -delayload:shlwapi.dll, \
     LIBS_unix := -ljvm -ljava $(LIBM), \
-    LIBS_linux :=  $(LIBDL), \
+    LIBS_linux :=  $(LIBDL) $(SPEECHD_LIBS), \
```
mkartashev: This adds a hard dependency to `libawt.so`, meaning that it won't be able to load without `libspeechd.so`. Previously, only backend-specific libraries had that dependency (like `libawt_xawt.so`). This way JDK could work in a headless system without `libspeechd.so`. I would prefer (and I think the OpenJDK community will, too) that this dependency is moved down to xawt and wlawt libraries.

mkartashev (changes requested):

make/modules/java.desktop/lib/Awt2dLibraries.gmk
tanya011: Could you please look one more time? But I'm still not sure

make/modules/java.desktop/lib/Awt2dLibraries.gmk
```diff
@@ -303,6 +291,7 @@ ifeq ($(call isTargetOs, windows macosx), false)
         include \
         common/awt/debug \
         common/awt/systemscale \
+        common/awt/AccessibleAnnouncer \
```
OnePatchGuy: I don't think we should expose the headers from that directory because they're only supposed to be used by that directory itself. I'd say `LIBAWT_XAWT_EXTRA_SRC` should be used instead (if needed at all).

make/modules/java.desktop/lib/Awt2dLibraries.gmk
```diff
@@ -315,10 +304,19 @@ ifeq ($(call isTargetOs, windows macosx), false)
...
 
+    # Setup speech-dispatcher a11y announcer for Linux xawt
+    ifeq ($(call isTargetOs, linux), true)
+      ifeq ($(A11Y_SPEECHD_ANNOUNCING_ENABLED), true)
+        A11Y_SPEECHD_ANNOUNCING_CFLAGS := $(SPEECHD_CFLAGS)
+      else
+        A11Y_SPEECHD_ANNOUNCING_CFLAGS := $(SPEECHD_CFLAGS) -DNO_A11Y_SPEECHD_ANNOUNCING
+      endif
+
+      LIBAWT_XAWT_CFLAGS += $(A11Y_SPEECHD_ANNOUNCING_CFLAGS)
```
OnePatchGuy: I feel like appending this unconditionally to avoid a sudden situation in the future when `-DNO_A11Y_SPEECHD_ANNOUNCING` doesn't accidentally get added.

make/modules/java.desktop/lib/Awt2dLibraries.gmk
```diff
@@ -408,6 +406,7 @@ ifeq ($(call isTargetOs, windows macosx), false)
         include \
         common/awt/debug \
         common/awt/systemscale \
+        common/awt/AccessibleAnnouncer \
```
OnePatchGuy: The same as for `LIBAWT_XAWT_EXTRA_HEADER_DIRS`

make/modules/java.desktop/lib/Awt2dLibraries.gmk
```diff
@@ -425,6 +424,17 @@ ifeq ($(call isTargetOs, windows macosx), false)
...
+    ifeq ($(call isTargetOs, linux), true)
+      ifeq ($(A11Y_SPEECHD_ANNOUNCING_ENABLED), true)
+        A11Y_SPEECHD_ANNOUNCING_CFLAGS_WLAWT := $(SPEECHD_CFLAGS)
+      else
+        A11Y_SPEECHD_ANNOUNCING_CFLAGS_WLAWT := $(SPEECHD_CFLAGS) -DNO_A11Y_SPEECHD_ANNOUNCING
+      endif
+
+      LIBAWT_WLAWT_CFLAGS += $(A11Y_SPEECHD_ANNOUNCING_CFLAGS_WLAWT)
+    endif
+
```
OnePatchGuy: It looks like exactly the same is done for XAWT. What if we move out the block to somewhere before both XAWT and WLAWT and set there toolkit-independent values to reuse in XAWT and WLAWT at the same time?

OnePatchGuy (changes requested):

make/modules/java.desktop/lib/Awt2dLibraries.gmk
tanya011: 8e40f332

make/modules/java.desktop/lib/Awt2dLibraries.gmk
tanya011: 3b7416a0

make/modules/java.desktop/lib/Awt2dLibraries.gmk
tanya011: 8e40f332

make/modules/java.desktop/lib/Awt2dLibraries.gmk
tanya011: 2e9c9459

mkartashev (approved):

make/modules/java.desktop/lib/Awt2dLibraries.gmk
OnePatchGuy, line 1: In the current state of the patch, it becomes impossible to disable the announcing feature via the `--with-speechd=no` autoconf flag.

For some reason, the moved files get compiled 4 times instead of two expected (once for _XAWT_ and once for _WLAWT_), and the other two times the compiler flags are different and don't include `-DNO_A11Y_SPEECHD_ANNOUNCING`. You can check this via running `make compile-commands` and then checking the file `compile_commands.json` in the build directory (just search for `AccessibleAnnouncer.c` for example).

I'm not sure why this happens, maybe because some other targets also include `src/java.desktop/unix/common/awt`, but in this case I'm surprised how something like that hasn't shot off our legs yet (cc @mkartashev).

If you wonder why those 2 additional compilations affect the build, my bet is most likely they belong to different shared libraries and when those get loaded, they kind of replace the symbol `Java_sun_swing_AccessibleAnnouncer_nativeAnnounce` by their own version (this mechanism is often used when you want to change some functionality of an app via [LD_PRELOAD](https://www.baeldung.com/linux/ld_preload-trick-what-is)). Or maybe it's some sort of [ODR violation](https://en.cppreference.com/w/cpp/language/definition.html#one_definition_rule), though it shouldn't exist in plain C AFAIU.

Anyway, the solution should have the following effects:
* `compile_commands.json` contains only 2 compiler invocations for `AccessibleAnnouncer.c`: one for XAWT (accompanied by a `-DXAWT` flag) and one for WLAWT (accompanied by a `-DWLAWT` flag)
* if `--with-speechd=no` is passed at the configure stage, the invocations should contain `-DNO_A11Y_SPEECHD_ANNOUNCING` in any case.

OnePatchGuy (changes requested):

make/modules/java.desktop/lib/Awt2dLibraries.gmk
OnePatchGuy: Addressed by [JBR-10004](https://youtrack.jetbrains.com/issue/JBR-10004)

make/modules/java.desktop/lib/Awt2dLibraries.gmk
mkartashev: Since the files are in the common part, they probably need to get explicitly excluded from compilation. See for example:
```
ifeq ($(DBUS_FOUND), false)
  LIBAWT_EXFILES += dbus_interface.c dbus_interface.h
endif
```

Not 100% sure this is *the* reason, though.

tanya011: @OnePatchGuy , both problems are solved: b6a35f9bfc4a22b9c2543ca439eccb5acfc4f28b

OnePatchGuy (approved):
Seems working fine, LGTM.

OnePatchGuy:
The patch has been rebased on top of _jbr25_ branch and merged into there and _main_ branches.

PR #598, opened 2026-04-09
JBR-10192 AccessibleAnnouncer works even when Orca is turned off

tanya011:
Detect running Orca process instead of checking "EnableSpeech" in the config file


OnePatchGuy:
Thanks! LGTM.

OnePatchGuy:
Merged into *jbr25* (commit `176b21cdd3be712d8a1f8a92c6c13a41f421d3c5`) and *main* branches.

PR #615, opened 2026-07-03
JBR-10347 Wayland: fix deadlock between WLClipboard dataLock and clipboard monitor

verglor:
Fixes [JBR-10347](https://youtrack.jetbrains.com/issue/JBR-10347) (also reported as [IJPL-248197](https://youtrack.jetbrains.com/issue/IJPL-248197)) — IDE-wide UI freezes/deadlocks on Wayland when a clipboard offer arrives while another thread reads the clipboard.

### Root cause

An AB–BA lock inversion, regressed from the [JBR-7896](https://youtrack.jetbrains.com/issue/JBR-7896) fix by the JBR-9898 change:

* **EDT**: `WLDataDevice.handleSelection` → `WLClipboard.handleClipboardOffer` takes **`dataLock`**, then `SunClipboard.lostOwnershipNow()` needs the **clipboard monitor** (`SunClipboard.java:242`)
* **Any thread**: `SunClipboard.getContents()` / `isDataFlavorAvailable()` (synchronized, holds the **monitor**) → `WLClipboard.getClipboardFormats()` needs **`dataLock`**

The JBR-7896 fix ("Avoid performing blocking I/O while holding a lock") deliberately kept `lostOwnershipNow()` outside `dataLock`, and the JBR-8833 refactor preserved that. JBR-9898's ownership check reads `dataLock`-guarded state (`ourDataSource`), so the `lostOwnershipNow()` call moved inside the synchronized block, re-creating the inversion. JBR-9865 shortened the freeze window (non-blocking `getMimes()`) but the inversion remained.

### Fix

Evaluate the ownership condition under `dataLock`, invoke `lostOwnershipNow()` after releasing it — restoring the JBR-7896 invariant of never entering clipboard-monitor code while holding `dataLock`.

### Test

`WLClipboardLockOrderTest` provokes the exact interleaving deterministically: a reader thread holds the clipboard monitor (as `SunClipboard.getContents()` does), waits for the offer-handler thread to block, then calls `getClipboardFormats()`. Failure is detected via `ThreadMXBean.findDeadlockedThreads()` and reported with the JVM deadlock report.

Verified on JBR 25.0.3+9-b329.124 under a real GNOME Wayland session:

* unpatched runtime → `RuntimeException: Deadlock detected` with stacks identical to the field thread dumps in JBR-10347
* same runtime with the fixed class via `--patch-module` → `Test passed`

### Backport

The `jbr25` tip carries the same inversion (with `lostOwnershipNow(null)` instead of `lostOwnershipNow()`, so not a clean cherry-pick). Since that line ships in 2026.1/2026.2 IDEs, I'm happy to submit the `jbr25` backport PR as well once this one is reviewed.


verglor:
Hi @vprovodin, is force pushing `main` branch regular workflow here ? It makes PRs based on it problematic - or should I use different branch as base for PRs ?

tsarn:
Hi,

Thank you for the pull request and your hard work. It saddens me to say it, but due to various legal reasons we cannot accept pull requests from external contributors: https://github.com/JetBrains/JetBrainsRuntime#contributing.

Consider contributing this to OpenJDK project Wakefield https://openjdk.org/projects/wakefield/ instead. Please also note that you cannot have used AI *at all* to help you develop this code. Otherwise it will be rejected by OpenJDK https://openjdk.org/legal/ai.

verglor:
Hi @tsarn, thank you for the information. So could you please fix it ? It seems to be trivial change and prevents serious deadlock.

PR #619, opened 2026-07-21
Add support for setting WS_EX_NOREDIRECTIONBITMAP on a new AWT window

m-sasha:
This PR adds a RootPane client property, `jbr.window.noRedirectionBitmap`, which sets the `WS_EX_NOREDIRECTIONBITMAP` flag on the window (in Windows).

This allows Skiko/Compose to overlay its own content such that window resizing is smooth (no white bars, no jumps).

https://github.com/user-attachments/assets/7d898412-39e6-4e8d-98a4-48ce4d0f4305

---------

### DIFF
--- a/src/java.desktop/windows/classes/sun/awt/windows/WWindowPeer.java
+++ b/src/java.desktop/windows/classes/sun/awt/windows/WWindowPeer.java
@@ -88,6 +88,9 @@ public class WWindowPeer extends WPanelPeer implements WindowPeer,
 
     public static final String WINDOW_CORNER_RADIUS = "apple.awt.windowCornerRadius";
 
+    // RootPane client property for setting WS_EX_NOREDIRECTIONBITMAP on the native window
+    public static final String WINDOW_NO_REDIRECTION_BITMAP = "jbr.window.noRedirectionBitmap";
+
     // we can't use WDialogPeer as blocker may be an instance of WPrintDialogPeer that
     // extends WWindowPeer, not WDialogPeer
     private WWindowPeer modalBlocker = null;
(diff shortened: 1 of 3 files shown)


src/java.desktop/windows/classes/sun/awt/windows/WWindowPeer.java
```diff
@@ -259,22 +262,32 @@ void initialize() {
             setOpaque(((Window)target).isOpaque());
         }
 
-        if (target instanceof RootPaneContainer) {
-            JRootPane rootpane = ((RootPaneContainer)target).getRootPane();
-            if (rootpane != null) {
-                setRoundedCornersImpl(rootpane.getClientProperty(WINDOW_CORNER_RADIUS));
-            }
+        setRoundedCornersImpl(getRootPaneClientProperty(WINDOW_CORNER_RADIUS));
```
avu, line 265: I see the change in behaviour - we call setRoundedCornersImpl unconditionally.

avu (changes requested):

avu (changes requested):
I think we need some robot test checking the behaviour.
Scenario:
 - resize, get color from the changing area, check for missing background

src/java.desktop/windows/classes/sun/awt/windows/WWindowPeer.java
m-sasha: True, but it ignores `null` values.

Do you want me to check for `null` before calling it anyway?

m-sasha:
> I think we need some robot test checking the behaviour. Scenario:
> 
> * resize, get color from the changing area, check for missing background

As far as I understand, this doesn't by itself fix resizing of windows in JBR. There's quite a lot of additional code in a skiko PR I'm working on to make it work. And it will only work for what skiko does: overlay a GPU surface on top of the window.

In fact, it makes the window content transparent.

image

m-sasha:
Converted to draft for now because I may have stumbled into a solution that doesn't require this.

m-sasha:
Looks like it's possible to make window resizing smooth without this flag.

PR #626, opened 2026-07-30
wayland: Add support for wl_fixes::ack_global_remove

zzag:
The wl_fixes.ack_global_remove request signals the compositor that the client will not bind the removed global. It can be used by the compositor to decide when it is safe to actually destroy the corresponding global. If a global is destroyed too soon, some clients may get disconnected.

---------


zzag:
See also https://blog.vladzahorodnii.com/2026/03/24/addressing-global-removal-race-in-wayland/ for more details about `wl_fixes.ack_global_remove`.

zzag:
`src/java.desktop/unix/native/libwlsplashscreen/splashscreen_sys.c` also gets a registry. I didn't port it to `wl_fixes.ack_global_remove` because splash screens are meant to be short-lived. Let me know if you'd prefer that code to ack `wl_registry.global_remove` events as well, but IIUC, it's not necessary.

zzag:
@tsarn @mkartashev I'm not entirely sure what the contributing process is. The CONTRIBUTING.md doc says that you need to contribute upstream, but wakefield is quite behind JBR. And IIUC the wayland changes in JBR are periodically backported to wakefield.

tsarn:
Hi, thanks for your contribution! Unfortunately, it's true that for legal reasons we can't accept code into JBR from external contributors. I've read your blog post a few months ago, and I was planning to implement this myself, once `wl_fixes` v2 is implemented in a popular compositor.

There's also a question of whether or not Wakefield folks would be okay with `#ifdef`ing this behavior, since it would subtly change, depending on which version of Wayland headers the JRE was compiled with. I personally think it might be okay, since it's just a bug fix, and not a feature, but maybe this is something worth discussing on the Wakefield mailing list?

You can contribute this directly to Wakefield, and we will merge this into JBR. Or since the change is small, I can reimplement this in JBR myself, and then it will get into Wakefield later.

Sorry for the mess that GitHub shows by the way, we rebase `main` nightly on top of OpenJDK's master. The `jbr25` branch is the branch which is stable and append-only.

zzag:
> once wl_fixes v2 is implemented in a popular compositor.

kwin 6.7 should support it. I don't know if Arch rebuilt kwin package to enable wl_fixes v2. If you build the `Plasma/6.7` branch in kwin, you should have wl_fixes v2.

> You can contribute this directly to Wakefield, and we will merge this into JBR. Or since the change is small, I can reimplement this in JBR myself, and then it will get into Wakefield later.

Contributing to OpenJDK is challenging because you need to file a task and bugzilla is very closed off for external contributors so I wouldn't mind if you reimplement this in JBR.

tsarn:
> kwin 6.7 should support it. I don't know if Arch rebuilt kwin package to enable wl_fixes v2. If you build the Plasma/6.7 branch in kwin, you should have wl_fixes v2.

Got it, thanks! My Fedora 44 machine has Plasma 6.7.3, but `wl_fixes` is still on version 1. I'll try building from source maybe.

zzag:
The relevant change is this one https://invent.kde.org/plasma/kwin/-/commit/8b8d2b24d1298dffd5f7c7f8cd384cce466752ea. It is likely that Fedora didn't rebuild kwin after wayland 1.26 got shipped. 6.7.4 should definitively support v2.

PR #628, opened 2026-08-05
JBR-10525 Add FramePacing service: display-aligned frame pacing clock

rock3r:
This adds a JBR API service delivering display-aligned ticks for rate-limiting UI work. The use case: a Compose for Desktop UI with a draw-only animation (e.g., a progress shimmer, in Studio's case) repaints as fast as the EDT allows and saturates the GPU — we measured an unpaced ComposePanel at ~1229 renders/sec and 14.7W of GPU on a 120Hz MacBook Pro (M1 Max); paced to the display it drops to ~121 renders/sec and 1.16W.

The API is based on display-scoped tick subscriptions, with queryable backend quality. v1 is a shared ESTIMATED backend — one refcounted timer thread per subscribed display, aligned to the refresh rate. Native clocks (CVDisplayLink, DWM composition timing) exist on a follow-up v2 branch and slot in behind the same surface as quality upgrades. V1 solves the immediate performance issue, by adding frame limiting, but not all the issues. V2 will also allow skiko to align to vsync to avoid tearing.

Note that the new service is inert by default; it needs to be explicitly enabled, like the test harness does. This design lets existing JBR usecases keep the existing performance profile, and lets users opt-in as they prefer. This design also lets consumer apps, like Android Studio, enable the limiter without having to wait for a Skiko or CMP change to be able to. I will open a CMP PR separately.

The service is in sun.awt, next to RoundedCornersManager and WindowMoveService. I chose to keep it close to similar services, but happy to move elsewhere as needed. The only change to an existing file is widening WLGraphicsDevice.getID()'s visibility: the registry id is the device's documented identity, and wl_output.name can be absent, so I do not think this could be done otherwise. Missed ticks are skipped, never queued; close() is non-blocking and safe from inside the listener. jtreg tests are under test/jdk/jb/FramePacing; validated on macOS, Windows and Linux.

Measurements, methodology and plan: https://frame-pacing.seeb.workers.dev — harness at https://github.com/rock3r/frame-pacing-test.

There is a companion PR on the JetBrainsRuntimeApi repository that adds the actual API for users that v2 will build upon: https://github.com/JetBrains/JetBrainsRuntimeApi/pull/31


bourgesl:
See #629, rebased on JBR main

rock3r:
Thanks @bourgesl!

bourgesl:
See my comments (your PR review) on https://github.com/JetBrains/JetBrainsRuntime/pull/629

bourgesl:
Looking forward the FramePacing v2 ...

rock3r:
I have a V2 draft but I need time to review it better and clean it up. I do have encouraging numbers tho from preliminary testing, but I'm unsure how reliable they are just yet. The branch is already on my fork if you're curious, but I'm prioritising making use of this API in Skiko to unblock the perf problem first. Then V2 is about correctness (vsync timing and actual frame pacing, especially with VRR displays, since a static timer can't do that)

rock3r:
Addressed comments from #629 (minus one, awaiting clarification), rebased on main

bourgesl (commented):
I will review last commit asap

rock3r:
Thank you. I have addressed the API side and rebased this on fresh main

src/java.desktop/share/classes/sun/awt/FramePacing.java
```diff
@@ -0,0 +1,329 @@
...
+            return clock.displayId;
+        }
+
+        @Override
+        public void close() {
+            synchronized (this) {
+                if (closed) return;
+                closed = true;
+            }
+            proxyPin = null;
```
bourgesl: move this cleanup after the unsubscribe that may use the listenerRef ?

src/java.desktop/share/classes/sun/awt/FramePacing.java
```diff
@@ -0,0 +1,329 @@
...
+
+    /**
+     * Called by a clock (from its tick thread's hotplug check) when its
+     * display disappeared. Removing the stopped clock from the registry is
+     * what makes display replug work: ids are commonly reused when the same
+     * monitor returns, and a stale entry would hand later subscribers a clock
+     * that never ticks. Closing the dead clock's remaining subscriptions
+     * afterwards is a harmless no-op.
+     */
+    private synchronized void displayGone(DisplayClock clock) {
```
bourgesl: maybe rename this method to removeClock(clock) that can be reused in few places:
- unsubscribe
- DisplayClock.run

bourgesl (changes requested):
Only 2 last minor comments.
LGTM

Congratulations!

rock3r:
All done! Thank you :)

bourgesl (approved):
LGTM

bourgesl:
@rock3r could you rebase your PR with today's main branch ?

bourgesl:
@avu could you have a look to the source code (LGTM) ?
If OK, how to push it ? should a new JBR ticket be created and used in the commit message ?

bourgesl:
Looking forward the PR for FramePacing v2 ...

rock3r:
I'll work on preparing it as soon as this is shipped :)

rock3r:
Just pushed because I found two races here while working on v2, so I've pushed a fix.

A clock stops itself from its own tick thread, without the service lock, and removes its registry entry in a separate step. That leaves two windows for `subscribe()`:

1. After the stop flag is set: `subscribe()` adopted the dying clock, so the caller got a `Subscription` that never ticked at all.
2. Before it's set: the tick thread had already emptied the listener list, and `add()` read that as "not started yet" and started a second tick thread for the same display.

The fix is simple: `subscribe()` now treats a stopped clock as missing; the clock tracks the "started" state explicitly instead of using empty list as a signal of it; and the do-I-stop decision moved into `retireIfIdle()`, so setting the flag and removing the entry happen together under the same lock.

`FramePacingSubscribeRaceTest` covers both. It plants the state instead of racing it, so it's deterministic.

rock3r:
Closing: this work has moved into Skiko.

The plan was a display-aligned pacing clock in the runtime, with Skiko and Compose Multiplatform as clients. While building the client we found the clock does not need runtime support: the same per-display vblank sources (CADisplayLink, `IDXGIOutput::WaitForVBlank`, DRM per-CRTC vblank) can be reached from Skiko using public AWT API only, with no `sun.awt` access.

On the Compose Multiplatform team's request we pivoted to a Skiko-only implementation, so that **every JVM benefits and not only the JetBrains Runtime**. That work is JetBrains/skiko#1282.

The measurements, the design reasoning and the history of this approach are kept at https://specs.sebastiano.dev/frame-pacing/

Nothing here was wrong; it simply became unnecessary. Reopening is cheap if a runtime-side clock is ever wanted again.

PR #631, opened 2026-08-18
JBR-10545 Fix round corner border geometry

stachenov:
Using RoundRectangle2D together with quadTo used for masks
doesn't seem to work well with high scaling levels.
Sharper quadratic shapes protrude beyond the border,
which is noticeable when the contrast between the border
and the background is high enough. For example,
in light themes with an almost white background
and a dark gray border.

Because RoundRectangle2D uses cubic curves under the hood,
the fix is using cubic curves for corners instead of quad curves.
Between this using quadTo for the border,
using cubic curves yields better-looking results.
With quadTo for the border, it looks somewhat off at the corners,
it's very clearly visible where lineTo ends and quadTo begins,
like if quadratic curves are thicker than straight lines.
Using cubic curves for everything looks way better.

The coefficients for cubic control points are calculated
using exactly the same logic as RoundRectangle2D (RoundRectIterator)
uses for its own shape. It's all based on a single value
which is multiplied by the arc radius to get control point coordinates.

---------

### DIFF
--- a/src/java.desktop/unix/classes/sun/awt/wl/WLWindowPeer.java
+++ b/src/java.desktop/unix/classes/sun/awt/wl/WLWindowPeer.java
@@ -483,27 +483,38 @@ public class WLWindowPeer extends WLComponentPeer implements SurfacePixelGrabber
         int h = getHeight();
         int radius = customCorners.radius();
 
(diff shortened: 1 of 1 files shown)


avu (approved):
LGTM

Issue #612, opened 2026-06-12
Severe performance regression between JBR jbr_jcef-21.0.9 and jbr_jcef-21.0.11 (problem still present in 25.0.3)


lagar84:
Hello,

On my java web application, startup of tomcat show a severe performance regression after going to 21.0.11+1-b1163.116 (from 21.0.9+1-b1038.78)

Startup on 21.0.9: `Server startup in [21220] milliseconds`

Startup on 21.0.11: `Server startup in [56071] milliseconds`

To isolate the problem, I created an empty demo application using https://start.spring.io/  (attached here, demo.zip , standard maven war).

[demo.zip](https://github.com/user-attachments/files/28879521/demo.zip)

Link of sample configuration if need to generate directly from start.spring.io : [link](https://start.spring.io/#!type=maven-project&language=java&platformVersion=4.0.7&packaging=war&configurationFileFormat=properties&jvmVersion=21&groupId=com.example&artifactId=demo&packageName=com.example.demo&dependencies=)

Logs of running in different JVMs:

[fast_jbr_21.0.9+1-b1038.78.txt](https://github.com/user-attachments/files/28879555/fast_jbr_21.0.9%2B1-b1038.78.txt)

[fast_temurin_openjdk_21.0.11+10-LTS.txt](https://github.com/user-attachments/files/28879558/fast_temurin_openjdk_21.0.11%2B10-LTS.txt)

[slow_jbr_21.0.11+1-b1163.116.txt](https://github.com/user-attachments/files/28879566/slow_jbr_21.0.11%2B1-b1163.116.txt)

Eclipse temurin (adoptium) and jbr_21.0.9: `Server startup in [4293] milliseconds` and `Server startup in [4254] milliseconds`

Jbr_21.0.11: `Server startup in [11379] milliseconds`

In the attached log files there is a complete tomcat log of each jvm run. All then running on default jvm parameters.

JBR-17.0.14+1-1367.22-jcef is also as fast as jbr_jcef-21.0.9

Please ping me if more information is needed.

Best wishes,

lagar84:
Running the spring demo app on jbr_jcef-25.0.3-windows-x64-b496.62 shows the same performance problem of Jbr_21.0.11 : `Server startup in [11223] milliseconds`

full log: 

[slow_jbr_25.0.3+9-b496.62.txt](https://github.com/user-attachments/files/28880566/slow_jbr_25.0.3%2B9-b496.62.txt)

[...]

PR #636, opened 2026-08-28
JBR-10418 improve file canonicalization performance

Azadios:
Replace expensive Path.toRealPath with cheaper Files.exists to probe directory existence.
Improve algorithmic complexity of resolving existing path prefix.

Fix for
https://github.com/JetBrains/JetBrainsRuntime/issues/612
https://youtrack.jetbrains.com/issue/JBR-10418

---------
- [X] I confirm that I make this contribution in accordance with the [OpenJDK Interim AI Policy](https://openjdk.org/legal/ai).


src/java.base/share/classes/java/io/IoOverNioFileSystem.java
```diff
@@ -568,29 +580,46 @@ private String canonicalize0(String path) throws IOException {
...
-
-                while (!nioPath.equals(nioPath.getRoot())) {
-                    Path parent = nioPath.getParent();
-                    if (parent == null) {
-                        parent = nioPath.getRoot();
+                Path root = nioPath.getRoot();
+                int totalPathElementCount = nioPath.getNameCount();
+
+                Path fullPath = resolveSubpath(root, nioPath, totalPathElementCount);
+                Path realPath = Files.exists(fullPath) ? toRealPathOrNull(fullPath) : null;
```
vladimirlagunov: Just `Path realPath = toRealPathOrNull(path)` 

If the file does not exist, `toRealPath()` throws `FileNotFoundException`, which is a subclass of `IOException`.

vladimirlagunov: Ouch, it's exactly that optimization.

Anyway, I'd avoid `Files.exists`. The problem is already defined in its docs:

> Note that the result of this method is immediately outdated. If this
> method indicates the file exists then there is no guarantee that a
> subsequent access will succeed. Care should be taken when using this
> method in security sensitive applications.

Probably, I have to run the benchmark.

Azadios:
Merged in https://github.com/JetBrains/JetBrainsRuntime/pull/640

